PR #30, opened 2023-09-29
Support for Intel cards.

Vladislav00:
The rssDepth parameter is now bound to the port to match the configuration of multiple boards with different functionality. added the rssDepth parameter without ipv6 extension support

### DIFF
--- a/dataplane/dataplane.cpp
+++ b/dataplane/dataplane.cpp
@@ -292,6 +292,190 @@ eResult cDataPlane::init(const std::string& binaryPath,
 	return result;
 }
 
+std::string rss_flags_to_string(uint64_t rss_flags)
+{
+	std::string flag_names;
+	if (rss_flags & RTE_ETH_RSS_IPV4)
+	{
+		flag_names += "IPV4 ";
+	}
+	if (rss_flags & RTE_ETH_RSS_FRAG_IPV4)
+	{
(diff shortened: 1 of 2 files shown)


dataplane/src/dataplane.cpp
```diff
@@ -1392,7 +1455,110 @@ eResult cDataPlane::parseJsonPorts(const nlohmann::json& json)
...
+				}
+			}
+		}
+		else
+		{
+			YADECAP_LOG_WARNING("not found: 'rssFlags', using default 'IP'\n");
+			rssFlags = RTE_ETH_RSS_IP;
+		}
+
+		config.ports[interfaceName] = {pci, rssFlags, bind};
```
taitov, line 1561: Let's make functions like `rss_flag_to_string()` and `string_to_rss_flag()`

dataplane/src/dataplane.cpp
```diff
@@ -342,21 +342,108 @@ eResult cDataPlane::initPorts()
 		rte_eth_conf portConf;
 		memset(&portConf, 0, sizeof(rte_eth_conf));
 
-		if (config.rssFlags != 0)
+		if (rssFlags != 0)
```
taitov, line 345: Rename to `rss_flags`

taitov (approved):

PR #34, opened 2023-10-03
GRE support naming refactoring

SteveBalayanAKAMedian:
The implementation of GRE tunneling scheme introduced updates to multiple structures in yanet. Yet the naming of the fields used for forwarding method in those structures was flawed. They were called 'tunnel', and that name could have led to confusion. 
This patch renames that field to 'forwarding_method'

### DIFF
--- a/common/balancer.h
+++ b/common/balancer.h
@@ -5,21 +5,21 @@
 namespace balancer
 {
 
-enum class tunnel : uint8_t
+enum class forwarding_method : uint8_t
 {
 	ipip,  // in services.conf it means lvs_method: TUN
 	gre,
 };
 
-constexpr const char* to_string(const tunnel& tunnel)
+constexpr const char* to_string(const forwarding_method& forwarding_method)
 {
-	switch (tunnel)
+	switch (forwarding_method)
 	{
-		case tunnel::ipip:
+		case forwarding_method::ipip:
 		{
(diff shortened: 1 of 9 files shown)


common/balancer.h
```diff
@@ -1,31 +1,30 @@
...
-	return "unknown";
-}
+    case forwarding_method::gre:
+	{
+        return "gre";
+    }
+    }
 
+    return "unknown";
+}
```
taitov, line 28: format

common/controlplaneconfig.h
```diff
@@ -309,7 +309,7 @@ using service_t = std::tuple<balancer_service_id_t,
                              std::optional<std::string>, ///< version
                              ::balancer::scheduler,
                              ::balancer::scheduler_params,
-							 ::balancer::tunnel,
+							 ::balancer::forwarding_method,
```
taitov, line 312: format

controlplane/src/balancer.cpp
```diff
@@ -735,7 +735,7 @@ void balancer_t::compile(common::idp::updateGlobalBase::request& globalbase,
 				flags,
 				counter_ids[0],
 			    	scheduler,
-					tunnel,
+					forwarding_method,
```
taitov, line 738: format

PR #40, opened 2023-10-04
Add more autotests

taitov:
Add autotests for check:
- checksum
- nat64stateless pools
- work with decap and nat module in one config
- nat64 + icmp
- route with mpls
- nat64stateless 1:N
- work nat64stateless with wkp or custom prefix


GeorgyKirichenko:
I do not like the last one commit, I would say the `sleep` was introduced for a reason - if one runs autotest-configured `dataplane` - it should not overheat the cpu.
But well, I believe you will return with a best solution.

PR #45, opened 2023-10-05
KNI is deprecated. Migrate to virtio_user

taitov:
https://doc.dpdk.org/guides/prog_guide/kernel_nic_interface.html
> KNI is deprecated and will be removed in future. See ABI and API Deprecation.
> 
> Virtio_user as Exception Path alternative is the preferred way for interfacing with the Linux network stack as it is an in-kernel solution and has similar performance expectations.


GeorgyKirichenko:
closes #45

GeorgyKirichenko (approved):

PR #55, opened 2023-10-13
Add support of IPv4/v6 outer source network for services

saushew:
Add the `ipv4_outer_source_network` and `ipv6_outer_source_network` parameters to services config in the controlplane.conf's balancer section. If these parameters are set, the netpacket source address sent from the load balancer will have the value corresponding to this parameter.

Fixes issue #60


dataplane/worker.cpp
```diff
@@ -3998,6 +3993,69 @@ inline void cWorker::balancer_tunnel(rte_mbuf* mbuf,
...
 }
 
+/// Sets the IPv6 source address for the packet, taking into account the address set for the service.
+/// In order to optimize the distribution of packets across the queues of the NIC installed on the servers
+/// to which the packet will be sent, the source address is (if possible) randomized taking into account the mask.
+inline void cWorker::set_ipv6_source(rte_ipv6_hdr* header,
+                                     const ipv6_address_t& balancer_src,
+                                     const dataplane::globalBase::ipv6_prefix_t& service_src,
+                                     const rte_ipv4_hdr* ipv4HeaderInner,
+                                     const rte_ipv6_hdr* ipv6HeaderInner)
```
taitov, line 4003: `set_ipv6_source()` - is not good name.
This is need only for balancer? If not, why one of argument is `balancer_src`?

dataplane/type.h
```diff
@@ -556,6 +556,26 @@ struct balancer_service_ring_t
 
 static_assert(YANET_CONFIG_COUNTERS_SIZE <= 0xFFFFFF, "invalid size");
 
+struct ipv4_prefix_t
```
taitov, line 559: move this struct to the same place, where dataplane::ipv4_address_t

dataplane/type.h
```diff
@@ -556,6 +556,26 @@ struct balancer_service_ring_t
...
+{
+	ipv4_prefix_t() :
+	        set(0)
+	{}
+
+	uint8_t set;
```
taitov, line 565: `set` looks like it is not part of ipv4_prefix_t

dataplane/type.h
```diff
@@ -556,6 +556,26 @@ struct balancer_service_ring_t
...
+	common::ipv4_prefix_t prefix;
```
taitov, line 566: in dataplane structs use `dataplane::ipv4_address_t` and `uint8_t mask` instead

dataplane/globalbase.cpp
```diff
@@ -325,16 +325,16 @@ eResult generation::get(const common::idp::getGlobalBase::request& request,
...
+	                return eResult::invalidInterfaceId;
+	        }
 
-		const auto& interface = interfaces[interfaceId];
+	        const auto& interface = interfaces[interfaceId];
 
-		std::get<2>(globalBaseResponse)[interfaceId] = {convert(interface.neighborEtherAddress),
-		                                                interface.flow};
+	        std::get<2>(globalBaseResponse)[interfaceId] = {convert(interface.neighborEtherAddress),
+	                                                        interface.flow};
```
taitov, line 337: revert please

taitov (approved):

PR #98, opened 2024-01-25
ci: rename jobs

taitov:


3Hren:
I think it was better. Simply because now we have to do brains-straining, remembering what name each Ubuntu has.

Issue #77, opened 2023-12-06
[DataPlane] Pure L3 load balancing


GeorgyKirichenko:
YANET load balancing works on L4 network level as it takes TCP/UPD ports into VS targeting and backend selection algo .

The idea is to implement pure L3 balancing - so the packet should be matched against a VS with destination IP only.

PR #114, opened 2024-02-06
Add support for pure L3 load balancing

yatancuyu:
#77


dataplane/worker.h
```diff
@@ -74,7 +74,7 @@ class cWorker
 
 	YANET_NEVER_INLINE void mainThread();
 
-	inline void calcHash(rte_mbuf* mbuf);
+	inline void calcHash(rte_mbuf* mbuf, bool without_ports = false);
```
yatancuyu, line 77: Is this acceptable?

dataplane/worker.cpp
```diff
@@ -4025,11 +4041,18 @@ inline void cWorker::balancer_handle()
 
 			key.balancer_id = metadata->flow.data.balancer.id;
 			key.protocol = metadata->transport_headerType;
+			key.l3_balancing = service.flags & YANET_BALANCER_PURE_L3;
 			key.addr_type = 6;
 			memcpy(key.ip_source.bytes, ipv6Header->src_addr, 16);
 			memcpy(key.ip_destination.bytes, ipv6Header->dst_addr, 16);
 			key.port_source = 0;
 			key.port_destination = 0;
```
yatancuyu, line 4049: if service requires pure L3 balancing, we leave ports fields with zeros, otherwise we miss session in balancer state in case of port changing. Hence we lose information about client_port and real_port in `yanet-cli balancer state`. Is this acceptable?

autotest/units/001_one_port/049_balancer_pure_l3/services.conf
```diff
@@ -0,0 +1,88 @@
+[
+  {
+    "vip": "10.0.0.16",
+    "proto": "tcp",
+    "scheduler": "rr",
+    "reals": [
```
yatancuyu, line 6: is is too implicit to declare pure L3 balancing service with just absence of `vport` field, should i have some flag for this?

GeorgyKirichenko: Looks good for me

dataplane/worker.cpp
GeorgyKirichenko: I do not think we should show ports here as many real TCP sessions will be mapped into one `balancer session`

dataplane/worker.h
GeorgyKirichenko: I definitely do not like the parameter naming. Also I would suggest to use a bitmask here like `L3_HASH | L4_HASH` and `L3_HASH` or something other like this.

libprotobuf/controlplane.proto
```diff
@@ -36,7 +36,7 @@ message BalancerRealFindRequest {
 message BalancerRealFindResponse {
   message Real {
     IPAddr ip = 1;
-    uint32 port = 2;
+    optional uint32 port = 2;
```
taitov, line 39: I think add flag `l3_balancing` is better

controlplane/configconverter.cpp
```diff
@@ -1666,13 +1666,13 @@ void config_converter_t::acl_rules_balancer(controlplane::base::acl_t& acl,
...
 				controlplane::base::acl_rule_transport_tcp_t rule_transport{range_t{0x0000, 0xFFFF},
-				                                                            vport};
+				                                                            vport.has_value() ? range_t{vport.value()} : range_t{0x0000, 0xFFFF}};
 				acl.nextModuleRules.emplace_back(rule_network, rule_transport, flow);
 			}
 			else if (proto == IPPROTO_UDP)
 			{
 				controlplane::base::acl_rule_transport_udp_t rule_transport{range_t{0x0000, 0xFFFF},
-				                                                            vport};
+				                                                            vport.has_value() ? range_t{vport.value()} : range_t{0x0000, 0xFFFF}};
```
taitov, line 1675: What about other protocols?

GeorgyKirichenko (commented):
I think it should be easy to enable other well-known protocols to L3 balancing.
We already have a protocol name into identifier mapping so there are two things left to do:
 1. Parse a protocol name into its identifier
 2. Handle new identifiers while emitting dispatcher rules 

However, configuration validation could be more complicated so I am Ok with merging this but filling an Issue for further work

GeorgyKirichenko (approved):

taitov (approved):

PR #120, opened 2024-02-08
[WIP] Dynamic allocation memory

taitov:


dataplane/neighbor.cpp
```diff
@@ -183,11 +191,11 @@ static void netlink_neighbor_dump(const std::function<void(const std::string&, c
 		return;
 	}
 
-	char buffer[8192];
-	int buffer_length = recv(nl_socket, buffer, sizeof(buffer), 0);
+	std::vector<char> buffer(64 * 1024);
```
GeorgyKirichenko, line 194: This does not look relevant to the PR subject

GeorgyKirichenko (commented):
I see new dataplane options but nothing of them are used for autotests

taitov:
> I see new dataplane options but nothing of them are used for autotests

In this PR (part 1) new options is useless =(
https://github.com/yanet-platform/yanet/pull/131 - part 2 with example in autotest

dataplane/updater.h
```diff
@@ -0,0 +1,219 @@
...
+	        pointer(nullptr)
+	{
+	}
+
+	eResult update(const std::vector<common::acl::tree_chunk_8bit_t>& values)
+	{
+		stats.extended_chunks_size = std::max((uint64_t)object_type::extended_chunks_size_min,
+		                                      values.size() / 16);
+
+		for (;;)
```
GeorgyKirichenko, line 59: So you just double the size of a LPM tree in case of no memory error?

dataplane/updater.h
```diff
@@ -0,0 +1,219 @@
...
+			                                              socket_id,
+			                                              object_type::calculate_sizeof(stats.pairs_size),
+			                                              stats.pairs_size);
+			if (pointer == nullptr)
+			{
+				return eResult::errorAllocatingMemory;
+			}
+
+			eResult result = pointer->fill(stats, values);
+			if (result != eResult::success)
```
GeorgyKirichenko, line 157: Do we really want to double the size of a hashtable in case of ANY error?

dataplane/updater.h
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	eResult update(const std::vector<common::acl::tree_chunk_8bit_t>& values)
+	{
+		stats.extended_chunks_size = std::max((uint64_t)object_type::extended_chunks_size_min,
+		                                      values.size() / 16);
+
+		for (;;)
+		{
+			pointer = memory_manager->create<object_type>(name.data(),
```
GeorgyKirichenko, line 61: I think, extended_chunks_size should be encapsulated into the tree but not the updater

dataplane/memory_manager.cpp
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	size += 2 * RTE_CACHE_LINE_SIZE;
+
+	YANET_LOG_INFO("yanet_alloc(name: '%s', socket: %u, size: %lu)\n",
+	               name,
+	               socket_id,
+	               size);
+
+	void* pointer = rte_malloc_socket(nullptr,
```
GeorgyKirichenko, line 86: I do not like the rte_malloc here as it could be fragile, for instance if we are near the overall allocation limits.

GeorgyKirichenko: At minimum a complex investigation in rte_malloc behavior should be done before its application for dynamic allocation purposes.
Personally, I am unsure that allocation sequences like A16 A16, A40 F16 F16 A40 (allocate 16GB twice, then reallocate the first one from 16 to 40 and then the second one 16 to 40) would be handled well.

dataplane/updater.h
taitov: No. Double size only if not enough chunks on update

dataplane/updater.h
taitov: The idea was that worker and lookup is not required extended_chunks_size

dataplane/updater.h
taitov: `fill` can return success, or not enough memory

dataplane/memory_manager.cpp
taitov: We will not make things worse in this place

GeorgyKirichenko (approved):

Issue #61, opened 2023-10-17
[DataPlane] Allow to configure keepstate timeouts


GeorgyKirichenko:
DataPlane emits a state record for each network packet matched against a firewall rule marked with a `keepstate` keyword.
Such state records are cleared up by GC after some period of inactivity configured `stateful_firewall_tcp_timeout` in case of a tcp session.
So there is no way to manage session lifetime depending either on its state (half-open, closed, etc) or source or destination.

There are following major directions to resolve the issue:
 - To implement more configuration options setting the GC behavior up, they could be configured statically using configuration file or dynamically using CLI/whatever else
 - To extend IPFW keepstate syntax with desired timeout of a state record to be emitted
 - To extend IPFW syntax with a non-terminating rule setting the default timeout for all following `keepstate` statement

However, all the directions do not contradict each other and may be implemented simultaneously.

PR #135, opened 2024-03-11
Configure firewall and balancer states lifetime options

GeorgyKirichenko:
The commit introduces two groups of state timeout options
The first one options group is for stateful firewall:
 - stateful_firewall_tcp_syn_timeout
 - stateful_firewall_tcp_syn_ack_timeout
 - stateful_firewall_tcp_fun_timeout
This options define the lifetime for TCP session depending on the last
processed session packet.

And the second group is for balancer states:
 - balancer_tcp_timeout
 - balancer_tcp_syn_timeout
 - balancer_tcp_syn_ack_timeout
 - balancer_tcp_fin_timeout
 - balancer_udp_timeout
 - balancer_other_protocols_timeout
All of them are defining timeouts for balancer sessions instead of
`balancer_state_ttl` which is obsoleted now.

Closes https://github.com/yanet-platform/yanet/issues/61


dataplane/worker_gc.cpp
```diff
@@ -927,10 +923,10 @@ void worker_gc_t::handle_free_mbuf()
 	}
 }
 
-inline bool worker_gc_t::is_timeout(const uint16_t timestamp,
-                                    const uint16_t timeout)
+inline bool worker_gc_t::is_timeout(const uint32_t timestamp,
+                                    const uint32_t timeout)
```
taitov, line 927: This change may break the GC for NAT, as there is a 16bit timestamp

dataplane/globalbase.cpp
```diff
@@ -338,16 +338,16 @@ eResult generation::get(const common::idp::getGlobalBase::request& request,
 	/** @todo
```
taitov, line 338: May be delete this comment block?

taitov (approved):

Issue #108, opened 2024-02-01
Holy crusade against copy-paste while parsing configs


3Hren:
# The problem
We have a bit of strange code in https://github.com/yanet-platform/yanet/blob/d53654af2450a6ed7630a4f6e6a2ac16abe50d17/dataplane/dataplane.cpp#L1772
Here, we parse config file and fill the https://github.com/yanet-platform/yanet/blob/d53654af2450a6ed7630a4f6e6a2ac16abe50d17/dataplane/dataplane.h#L331 using the following pattern:

```c++
if (exist(json, "some_key")) {
    configValues[eConfigType::some_enum] = json["some_key"];
}
```

, but such pattern is prone to copy-paste errors. What if we slightly change the key and check for existing for another one?

I suggest to:
1. Replace this pattern to a single call using https://json.nlohmann.me/api/basic_json/value/
Then, the code above will transform to something like:
```c++
configValues[eConfigType::some_enum] = json.value("some_key", 0); // Or meaningful default value.
```
2. Replace the `map<enum, uint64>` to a named struct.
3. Decompose it into a separate file.
4. Write an ADL deserializer nearby.
5. Then, we won't need for https://github.com/yanet-platform/yanet/blob/d53654af2450a6ed7630a4f6e6a2ac16abe50d17/dataplane/dataplane.h#L127, and can access config values directly.

We can start with dataplane config, but, of course, this can be propagated further to other configs.

# The profit
- Less copy-paste.
- Slightly faster access, as there is no need to recheck while accessing values. More precisely, this function will be removed:
```c++
uint64_t cDataPlane::getConfigValue(const eConfigType& type) const {
	if (configValues.find(type) == configValues.end()) {
		YADECAP_LOG_ERROR("unknown variable\n");
		return 0;
	}

	return configValues.find(type)->second;
}
```

# Alternatives
- Leave as is.

PR #137, opened 2024-03-19
Store config in struct instead of map

TheRandomCharacter:
#108

### DIFF
--- a/dataplane/config_values.h
+++ b/dataplane/config_values.h
@@ -1,0 +1,110 @@
+#pragma once
+
+#include<common/config.h>
+#include <nlohmann/json.hpp>
+
+#include <cstdint>
+
+struct ConfigValues {
+	uint64_t port_rx_queue_size = 4096;
+	uint64_t port_tx_queue_size = 4096;
+	uint64_t ring_highPriority_size = 64;
+	uint64_t ring_normalPriority_size = 256;
+	uint64_t ring_lowPriority_size = 64;
+	uint64_t ring_toFreePackets_size = 64;
+	uint64_t ring_log_size = 1024;
+	uint64_t fragmentation_size = 1024;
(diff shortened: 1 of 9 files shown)


dataplane/config_values.h
```diff
@@ -0,0 +1,108 @@
...
+	uint64_t balancer_tcp_timeout = YANET_CONFIG_BALANCER_STATE_TIMEOUT_DEFAULT;
+	uint64_t balancer_tcp_syn_timeout = YANET_CONFIG_BALANCER_STATE_TIMEOUT_DEFAULT;
+	uint64_t balancer_tcp_syn_ack_timeout = YANET_CONFIG_BALANCER_STATE_TIMEOUT_DEFAULT;
+	uint64_t balancer_tcp_fin_timeout = YANET_CONFIG_BALANCER_STATE_TIMEOUT_DEFAULT;
+	uint64_t balancer_udp_timeout = YANET_CONFIG_BALANCER_STATE_TIMEOUT_DEFAULT;
+	uint64_t balancer_other_protocols_timeout = YANET_CONFIG_BALANCER_STATE_TIMEOUT_DEFAULT;
+};
+
+inline void from_json(const nlohmann::json& j, ConfigValues& cfg) {
+	#define from_json_if_exists(some_key) \
```
3Hren, line 44: This can be the place where we accidentally break API in case of some field renaming. If so - such values from old JSONs will be no longer valid, and we silently ignore them, falling back to default values.
I'd rather sacrifice code bloat here in the name of clarity at this point.

3Hren (changes requested):
The other stuff LGTM.

TheRandomCharacter:
Decoupled configuration names from structure field names.
Complied to .clang-format

PR #155, opened 2024-04-04
Change the use of rte_ring for dumping in/out/drop packets to a newer…

yevzman:
… approach


dataplane/dataplane.cpp
```diff
@@ -1327,6 +1348,28 @@ eResult cDataPlane::splitSharedMemoryPerWorkers()
...
+			const auto units_number = getConfigValues().ring_lowPriority_size;
+			const auto size = sizeof(sharedmemory::ring_header_t) + unit_size * units_number;
+			ring.init(memaddr, unit_size, units_number);
+			offsets[shm] += size;
+
+			worker->lowPriorityRing = ring;
+
+			auto meta = common::idp::get_shm_info::dump_meta(name, "lp", unit_size, units_number, core_id, socket_id, key, offset);
+			dumps_meta.emplace_back(meta);
+		}
```
ol-imorozko, line 1372: This almost looks like a previous block of code (body of `for (const auto& [tag, ring_cfg] : config.shared_memory)` loop).
I think we could extract common functionality to a function.
something like:
```cpp
  void foo(arguments)
  {
      auto name = base_name + std::to_string(core_id) + (is_low_priority ? "" : "_" + std::to_string(ring_id));
  
      auto unit_size = sizeof(sharedmemory::item_header_t) + dump_size;
      if (unit_size % RTE_CACHE_LINE_SIZE != 0)
      {
          unit_size += RTE_CACHE_LINE_SIZE - unit_size % RTE_CACHE_LINE_SIZE;
      }
  
      auto size = sizeof(sharedmemory::ring_header_t) + unit_size * units_number;
      auto offset = offsets[shm];
      auto memaddr = reinterpret_cast<void*>(reinterpret_cast<std::uintptr_t>(shm) + offset);
  
      sharedmemory::cSharedMemory ring;
      ring.init(memaddr, unit_size, units_number);
      offsets[shm] += size;
  
      auto meta = common::idp::get_shm_info::dump_meta(name, tag, unit_size, units_number, core_id, socket_id, key, offset);
      dumps_meta.emplace_back(meta);
  }
```
as we cannot interact with private `worker` fields in the regular function.

Then the "low priority ring" turns into:
```
foo(...);
worker->lowPriorityRing = ring;
```

and loop body turns into
```
foo();
worker->dumpRings[ring_id] = ring;                                                                                                                                  
tag_to_id[tag] = ring_id;                                          
ring_id++;  
```

dataplane/dataplane.cpp
```diff
@@ -1327,6 +1348,28 @@ eResult cDataPlane::splitSharedMemoryPerWorkers()
 
 			ring_id++;
 		}
+		// init lowPriority ring
+		{
+			auto name = "r_lp_" + std::to_string(core_id);
+			auto offset = offsets[shm];
+			auto memaddr = (void*)((intptr_t)shm + offset);
```
ol-imorozko, line 1355: Avoid c-style casts.
It's generally better to avoid casts whatsoever, but I understand that some things cannot be done without them (though I'm pretty sure we can avoid some of them by using templates and/or `std::variant`, but that's a topic for discussion).

Here we should use named casts: `auto memaddr = reinterpret_cast<void*>(reinterpret_cast<std::uintptr_t>(shm) + offset);`

ol-imorozko (changes requested):

dataplane/dataplane.cpp
```diff
@@ -1327,6 +1348,28 @@ eResult cDataPlane::splitSharedMemoryPerWorkers()
...
+			sharedmemory::cSharedMemory ring;
+
+			auto unit_size = sizeof(sharedmemory::item_header_t) + MAX_PACK_SIZE;
+			if (unit_size % RTE_CACHE_LINE_SIZE != 0)
+			{
+				unit_size += RTE_CACHE_LINE_SIZE - unit_size % RTE_CACHE_LINE_SIZE; /// round up
+			}
+			const auto units_number = getConfigValues().ring_lowPriority_size;
+			const auto size = sizeof(sharedmemory::ring_header_t) + unit_size * units_number;
+			ring.init(memaddr, unit_size, units_number);
```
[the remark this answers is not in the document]
GeorgyKirichenko: Good catch

dataplane/dataplane.cpp
GeorgyKirichenko: We use C-style typecast in sake of performance and the code is on our most important hot-path.

Also this memory region will contain many data structures (and arrays of them) mapped dynamically with data offsets. Another requirement that the region layout should be parseable by third-party application written in C or Go for example.

However, I think we could improve the approach but let us do this in the another piece of work.

ol-imorozko: Okay, but I cannot resolve the conversation by myself, I guess I don't have the related permissions :)

TheRandomCharacter: > Unlike static_cast, but like const_cast, the reinterpret_cast expression does not compile to any CPU instructions (except when converting between integers and pointers, or between pointers on obscure architectures where pointer representation depends on its type). It is primarily a compile-time directive which instructs the compiler to treat expression as if it had the type target-type.

So there is no penalty in using `reinterpret_cast`. It also has a benefit of being easily greped

GeorgyKirichenko:
superseded by !259

PR #161, opened 2024-04-08
Switch to using arrays instead of maps for KNIs

TheRandomCharacter:


dataplane/burst.h
```diff
@@ -0,0 +1,137 @@
...
+
+		return *this;
+	}
+};
+
+template<uint16_t BURST_SIZE>
+class BasicBurst
+{
+	using iterator = rte_mbuf**;
+	rte_mbuf* mbufs_[BURST_SIZE];
```
ol-imorozko, line 83: `std::array<rte_mbuf*, BURST_SIZE> mbufs_;`

ol-imorozko (changes requested):

dataplane/base.h
```diff
@@ -14,13 +14,48 @@
...
 {
 
+class PortMapper
+{
+	static constexpr tPortId INVALID_PORT_ID = std::numeric_limits<tPortId>::max();
+	uint16_t ports_count_ = 0;
+	tPortId dpdk_ports_[CONFIG_YADECAP_PORTS_SIZE] = {INVALID_PORT_ID}; // logical to dpdk
+	tPortId logical_ports_[std::numeric_limits<tPortId>::max() + 1] = {INVALID_PORT_ID}; // dpdk to logical
+public:
+	uint16_t size() const { return ports_count_; }
```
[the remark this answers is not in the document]
TheRandomCharacter: You might have noticed that this function's name does not start with a capital letter,  in contrast to most other functions . This is supposed to convey that behavior of this function is analogous to STL. Marking this function `[[nodiscard]]` makes it differ. If you are reciting a commandment, breaking which is a mortal sin, then we should also change the name. What would be a good substitute? `length()`?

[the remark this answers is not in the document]
ol-imorozko: No, I mean, it's perfectly fine as-is if the goal is to create a function that is analogous to those in the STL

dataplane/base.h
```diff
@@ -14,13 +14,48 @@
...
+		else
+		{
+			YANET_LOG_ERROR("CONFIG_YADECAP_PORTS_SIZE exceeded");
+			return {};
+		}
+	}
+	tPortId ToDpdk(tPortId logical) const { return logical < CONFIG_YADECAP_PORTS_SIZE ? dpdk_ports_[logical] : INVALID_PORT_ID; }
+	tPortId ToLogical(tPortId dpdk) const { return logical_ports_[dpdk]; }
+	bool ValidDpdk(tPortId dpdk) const { return logical_ports_[dpdk] != INVALID_PORT_ID; }
+	bool ValidLogical(tPortId logical) const { return logical < INVALID_PORT_ID; }
```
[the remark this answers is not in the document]
TheRandomCharacter: What will be the meaning of `[[nodiscard]]` here? Do we require an action from the caller? Do we return an error that needs to be handled? Do we return a unique handle to something that is lost forever if the caller discards it?

dataplane/base.h
```diff
@@ -14,13 +14,48 @@
 namespace dataplane::base
 {
 
+class PortMapper
+{
+	static constexpr tPortId INVALID_PORT_ID = std::numeric_limits<tPortId>::max();
+	uint16_t ports_count_ = 0;
+	tPortId dpdk_ports_[CONFIG_YADECAP_PORTS_SIZE] = {INVALID_PORT_ID}; // logical to dpdk
```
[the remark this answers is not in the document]
TheRandomCharacter: Preferring `std::array` here is a valid point based on the fact that this array's size is  refered to later in the code.

dataplane/base.h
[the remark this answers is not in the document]
ol-imorozko: No, I'm just used to putting `[[nodiscard]]` on such small functions that return a certain result.
But I agree that it may be redundant

dataplane/base.h
```diff
@@ -14,13 +14,48 @@
...
+	tPortId logical_ports_[std::numeric_limits<tPortId>::max() + 1] = {INVALID_PORT_ID}; // dpdk to logical
```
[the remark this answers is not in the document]
TheRandomCharacter: Using `std::array` is moot here. We solely use elements access without bounds checking.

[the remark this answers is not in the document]
ol-imorozko: But this isn't about boundary checking.
In my opinion, it’s simply better to use `std::array` instead of C-style arrays.
After all, we are writing C++ code, and `std::array` does not compromise on performance.

dataplane/base.h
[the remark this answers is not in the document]
TheRandomCharacter: IMHO, blanketly using `[[nodiscard]]` detracts from the message that this functions(specifically) return value should not be discarded. In general any functions return value should not be discarded, but this functions -- especially.

There seems to be no consensus on the matter even in C++20 STL: for `vector` `empty()` is `[[nodiscard]]` while `size()` is not.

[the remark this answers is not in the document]
ol-imorozko: I think we can resolve this conversation, as it is a topic for debate. Microsoft, for example, marked _a lot_ of their functions in the STL as [[nodiscard]]: https://github.com/microsoft/STL/issues/206

When we adopt a specific coding standard, we could specify what actions to take regarding [[nodiscard]]. A linter like clang-tidy could easily mark them if we agree on adding them, or remove them if we decide otherwise.

I'm not able to resolve the conversation myself; I guess I don't have the required permissions yet. :)

dataplane/burst.h
```diff
@@ -0,0 +1,125 @@
...
+{
+	rte_mbuf** first_;
+	uint16_t len_ = 0;
+
+public:
+	BurstView(rte_mbuf** first, uint16_t length) :
+	        first_{first}, len_{length}
+	{
+	}
+	bool empty() { return len_ == 0; }
```
ol-imorozko, line 22: Here we should put `[[nodiscard]]` btw like the STL does.
(they have a valid reasoning, as new users tend to think that it's a verb, and that it will remove all the elements of a container).

TheRandomCharacter: Good insight, was wondering why they marked it and not `size()`

dataplane/burst.h
```diff
@@ -0,0 +1,124 @@
...
+	BurstSpan& Tx(tPortId port_id, tQueueId queue_id)
+	{
+		if (empty())
+			return *this;
+		auto txSize = rte_eth_tx_burst(port_id, queue_id, first_, len_);
+		len_ -= txSize;
+		first_ += txSize;
+		return *this;
+	}
+	BurstSpan& Tx(tPortId port_id, tQueueId queue_id, uint64_t& opackets, uint64_t& obytes)
```
GeorgyKirichenko, line 59: This may be confusing as the function does not "return" `opackets` and `obytes` but increases them

GeorgyKirichenko (commented):
I couldn't figure out whether the Burst object becomes the owner of a `mbuf` or not.
So the questions are:
 - should `clear` free all contained mbufs? If they are freed somewhere else before the `clear` then BaseBurst (and originated Span) turns to an invalid state. In the other hand if we call `clear` first then someone else should contain all this `mbufs` what makes baseBurst pointless here.
 - BurstSpan calls `tx_burst` and all sucessfully processed mbufs will be freed inside `DPDK` whereas failed ones will remain in the memory and owning BaseBurst turns to an invalid state.

dataplane/burst.h
```diff
@@ -0,0 +1,124 @@
...
+
+		len_ -= txSize;
+		first_ += txSize;
+
+		bytes -= Bytes();
+		obytes += bytes;
+
+		return *this;
+	}
+};
```
ol-imorozko, line 77: It really looks like one method can be used in another to avoid code duplication, but I can't think of a solution on top of my head.
The first idea is to use smth like:
```cpp
private:
BurstSpan& Tx(tPortId port_id, tQueueId queue_id, uint64_t* opackets, uint64_t* obytes) {
               if (empty())
			return *this;
		uint64_t bytes = Bytes();
		auto txSize = rte_eth_tx_burst(port_id, queue_id, first_, len_);
		len_ -= txSize;
		first_ += txSize;
		
		if (opackets != nullptr)
		 *opackets += txSize;
		
		if (obytes != nullptr) {
		bytes -= Bytes();
		*obytes += bytes
		}
		return *this
}
public:
BurstSpan& Tx(tPortId port_id, tQueueId queue_id) {
return Tx(port_id, queue_id, nullptr, nullptr);
}
BurstSpan& Tx(tPortId port_id, tQueueId queue_id, uint64_t& opackets, uint64_t& obytes) {
return Tx(port_id, queue_id, &opackets, &obytes);
}
```

dataplane/burst.h
```diff
@@ -0,0 +1,124 @@
...
+		if (len_ < mbufs_.size())
+		{
+			mbufs_[len_++] = m;
+			return true;
+		}
+		else
+		{
+			return false;
+		}
+	}
```
ol-imorozko, line 106:
```cpp
bool can_push = len_ < mbufs_.size();
if (can_push)
   PushUnsafe(m);
return can_push;
```

dataplane/burst.h
```diff
@@ -0,0 +1,124 @@
...
+	BurstSpan Span()
+	{
+		return BurstSpan{mbufs_.data(), len_};
+	}
+};
+
+using Burst = BasicBurst<CONFIG_YADECAP_MBUFS_BURST_SIZE>;
+
+} // namespace dpdk
```
ol-imorozko, line 124: EOF

dataplane/burst.h
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}
+	void PushUnsafe(rte_mbuf* m)
+	{
+		mbufs_[len_++] = m;
+	}
+	BurstSpan Rx(tPortId port_id, tQueueId queue_id)
+	{
+		len_ = rte_eth_rx_burst(port_id, queue_id, mbufs_.data(), mbufs_.size());
+		return BurstSpan{mbufs_.data(), len_};
```
ol-imorozko, line 114: just `return Span()`

dataplane/controlplane.cpp
```diff
@@ -509,27 +507,36 @@ common::idp::getControlPlanePortStats::response cControlPlane::getControlPlanePo
...
-				(void)coreId;
+				const auto& stats = kernel_stats[i];
+				const auto& portId = portmapper.ToDpdk(i);
 
-				controlPlane_drops += worker->statsPorts[portId].controlPlane_drops;
-			}
+				uint64_t controlPlane_drops = 0;
+				for (const auto& [coreId, worker] : dataPlane->workers)
+				{
+					(void)coreId;
```
ol-imorozko, line 520: Structured bindings are not variables, they cannot be "used".
I suggest removing this line as this accomplishes nothing.
Also it's c-style cast, which is generally a good thing to avoid.

Or don't, I think it will be fixed anyway in bulk when we adopt some coding standard and set up linters

ol-imorozko (changes requested):

dataplane/burst.h
TheRandomCharacter: Rewrote `Burst`, ditched `BurstSpan`

dataplane/base.h
```diff
@@ -14,13 +14,48 @@
...
+				YANET_LOG_ERROR("Duplicate dpdk port id provided to PortMapper");
+				return {};
+			}
+		}
+		else
+		{
+			YANET_LOG_ERROR("CONFIG_YADECAP_PORTS_SIZE exceeded");
+			return {};
+		}
+	}
```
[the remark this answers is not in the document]
TheRandomCharacter: IMHO nesting provides better clarity. Maybe we should discuss a threshold?

dataplane/burst.h
TheRandomCharacter: Suggested solution adds branching on the packet processing pass. This branching is superfluous, considering that decision to count statistics is actually made at compile time.

ol-imorozko: Yes, I know it's not ideal, it was just my first thought to show that we _can_ avoid code duplication.
One _possible_ way is do smth like that:
```cpp
template<typename TPacketPtr = nullptr_t, typename TBytePtr = nullptr_t>
BurstSpan& Tx(tPortId port_id, tQueueId queue_id, TPacketPtr opackets = nullptr, TBytePtr obytes = nullptr) {
    if (empty())
        return *this;
    
    uint64_t bytes = Bytes();
    auto txSize = rte_eth_tx_burst(port_id, queue_id, first_, len_);
    len_ -= txSize;
    first_ += txSize;
    
    if constexpr (!std::is_same_v<TPacketPtr, nullptr_t>) {
        *opackets += txSize;
    }
    
    if constexpr (!std::is_same_v<TBytePtr, nullptr_t>) {
        bytes -= Bytes();
        *obytes += bytes;
    }
    
    return *this;
}

// For cases where opackets and obytes are not used:
BurstSpan& Tx(tPortId port_id, tQueueId queue_id) {
    return Tx<>(port_id, queue_id);
}

// For cases where opackets and obytes are used:
BurstSpan& Tx(tPortId port_id, tQueueId queue_id, uint64_t& opackets, uint64_t& obytes) {
    return Tx(port_id, queue_id, &opackets, &obytes);
}
```

But, I mean, maybe there's a way to just call a smaller function from the bigger one and utilize its return value. I just didn't think about that enough to come up with a good solution.

dataplane/base.h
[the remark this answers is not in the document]
ol-imorozko: It's generally easier to follow the code when we handle edge cases first and get them out of the way.
This way, the "happy path" stands out without being buried in nested ifs.
It's like "If anything's wrong, let's deal with it right away; otherwise, here's what we're really here to do."

But of course it's not that important, so you can leave it as it is if you want.

However, choosing max column width is a great idea (we will choose it when we adopt some coding standard, of course)

TheRandomCharacter:
Threw away BurstSpan and rewrote Burst.

dataplane/controlplane.cpp
ol-imorozko: Oh, yea, now I see. Sure, we'll modify such things later

dataplane/burst.h
ol-imorozko: I see that you added TxTracked, I like it.
(I'm still not able to close this conversation by myself though, could you close it for me?)

TheRandomCharacter: Reworked without output parameters

dataplane/base.h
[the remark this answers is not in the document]
TheRandomCharacter: Let's discuss this later, while choosing codestyle

TheRandomCharacter:
Divided into two PRs

PR #164, opened 2024-04-11
Feature: Implement Support for Enhanced ACL Counters

saushew:
This pull request aims to enhance `YANET's` ACL counters feature by implementing support for a new type of rule syntax inspired by `IPFW`. The proposed enhancement allows for more accurate tracking of active rules by associating counters with specific rule entries.

The current implementation of ACL counters in `YANET` has a notable limitation where intersecting rules may lead to false positive conclusions about rule activity. This limitation arises due to the top-down packet firewall rule lookup process, where only the counter of the terminating rule is incremented.

To address this issue, this pull request introduces support for rules of the form `add count:counter_name`, allowing administrators to specify a counter name associated with each rule. By associating counters directly with rules, we ensure that counters accurately reflect rule activity, even in cases of rule intersections.

**Key Changes:**

- Introduce new rule syntax: `add count:counter_name`
- Update rule processing logic to support associated counters
- Implement necessary adjustments to maintain compatibility with existing rule sets


ol-imorozko:
I don't know if we do this, but could you add this description to commit message too?

GeorgyKirichenko:
I like the idea of tagging dumps and customizable counters but there are some points about:
 - I think that colon symbol does not fit well into `ipfw` syntax despite the fact that we agreed to extend them
 - We working on generalization of non-terminating rule processing (like dump, log, count, check-state, timeouts and others)
 
So let us to set the PR on hold at least until the non-terminating rules design makes clear.

GeorgyKirichenko:
This was superseded by another PR

PR #165, opened 2024-04-11
The first part of the changes: The transfer of counters to Shared memory

stal76:


common/pde.h
```diff
@@ -0,0 +1,658 @@
+#pragma once
+
+#include <rte_build_config.h>
```
GeorgyKirichenko, line 3: We should not rely on DPDK elsewhere except DataPlane.
In the opposite case this means that any utility as `cli` should depended on DPDK and be linked with.

common/pde.h
```diff
@@ -0,0 +1,658 @@
...
+		for (auto [core_id, socket_id] : workers_gc_to_sockets)
+		{
+			workers_gc[core_id].buffer = PtrAdd(sockets[socket_id], workers_gc[core_id].start_at_memory);
+		}
+
+		return eResult::success;
+	}
+};
+
+struct MainFileData
```
GeorgyKirichenko, line 512: I feel the name could be misleading. The PR suggests that only `DataPlane` manages counters in shared memory what could not be true in the future.

common/pde.h
GeorgyKirichenko, line 1: I would ask you to annotate classes and methods with commentaries denoting its purpose as well as overall memory layout.

GeorgyKirichenko (changes requested):
Thanks for the PR, I feel it could improve YANET as well.
However, please find some commentaries bellow:
 - Please split the PR into three commits (or even different PRs):
   - Minor fixes (Log output and e.g.)
   - Bus counters extension
   -  Shared memory and counters
 - The last one should be commented both in the commit message and in the code. I would like to see here motivation, overall design, memory layouts, class and methods purposes and descriptions
 - I feel that two-step `DataPlane` shared counters data initialization is dangerous and would prefer something like `dataPlaneData = sharedData::DataPlaneRead()` and further read of the values.
 - Also there should be an interface to determine if huge page memory should be used before the memory access.
- I did not understand how the mapping correctness could be validated, so the question is: what would be happen if `DataPlane` changes counter mapping while someone read the same memory chunk. I would suggest the "generation approach":
   - `DataPlane` maintains the generation value which changes every time if counters layout is changed.
   - Reader compares two generation values before and after the data read and retries if there is a discrepancy between them.

GeorgyKirichenko:
Closed as already implemented

PR #170, opened 2024-04-13
Compilation fixes

GeorgyKirichenko:
I discovered this errors during `gcc` and `clang` version investigation

### DIFF
--- a/dataplane/worker.cpp
+++ b/dataplane/worker.cpp
@@ -4413,7 +4413,7 @@ inline void cWorker::balancer_ipv6_source(rte_ipv6_hdr* header,
 		random_ptr++;
 	}
 
-	uint8_t n_to_fill = RTE_MIN(sizeof(header->src_addr) - random_ptr, 4);
+	uint8_t n_to_fill = RTE_MIN(sizeof(header->src_addr) - random_ptr, (size_t)4);
 
 	/// Fill the first 4 (or less) bytes after the mask with random numbers.
 	rte_memcpy(&header->src_addr[random_ptr], &random_src, n_to_fill);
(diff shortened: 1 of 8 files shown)


dataplane/worker.cpp
```diff
@@ -4413,7 +4413,7 @@ inline void cWorker::balancer_ipv6_source(rte_ipv6_hdr* header,
 		random_ptr++;
 	}
 
-	uint8_t n_to_fill = RTE_MIN(sizeof(header->src_addr) - random_ptr, 4);
+	uint8_t n_to_fill = RTE_MIN(sizeof(header->src_addr) - random_ptr, (unsigned long)4);
```
3Hren, line 4416: What's the problem with it?
Also `std::size_t` instead of `unsigned long`, because `decltype(sizeof(...))` is `size_t`.

3Hren (changes requested):

dataplane/worker.cpp
GeorgyKirichenko: gcc 13.2.1 bothers about the comparison
```
../dataplane/worker.cpp: In member function ‘void cWorker::balancer_ipv6_source(rte_ipv6_hdr*, const ipv6_address_t&, const dataplane::globalBase::balancer_service_t&, const rte_ipv4_hdr*, const rte_ipv6_hdr*)’:
../subprojects/dpdk/lib/eal/include/rte_common.h:617:20: error: comparison of integer expressions of different signedness: ‘long unsigned int’ and ‘int’ [-Werror=sign-compare]
  617 |                 _a < _b ? _a : _b; \
      |                 ~~~^~~~
../dataplane/worker.cpp:4397:29: note: in expansion of macro ‘RTE_MIN’
 4397 |         uint8_t n_to_fill = RTE_MIN(sizeof(header->src_addr) - random_ptr, 4);
```

3Hren: Okay. But the `typeof` of `sizeof` operator is implementation defined, but is `size_t`. So instead of `unsigned long` it's better to use `[std::]size_t`.

3Hren (approved):

PR #174, opened 2024-04-15
WIP: Dpdk bundled

GeorgyKirichenko:
Use statically built bundled version of `DPDK` with our patchs included


GeorgyKirichenko:
Also I have to clean the Docker images up

PR #175, opened 2024-04-15
Add a wrapper to encapsulate DPDK burst

TheRandomCharacter:


3Hren:
Please explain what is the purpose of this class? Where will it be used?

3Hren (changes requested):
Please explain what is the purpose of this class? Where will it be used?

TheRandomCharacter:
When working with dpdk bursts directly implies that burst length while being an integral part of burst is managed manually. This wrapper is intended to bundle together dpdk bursts cstyle array of pointers to packets with it's length and provide an interface that manages said length automatically.

This approach removes duplicate code of tracking number of packets that are currently in the burst and replaces it with a single implementation in the wrapper.

I would like us to use this wrapper as a preferred way of working with dpdk bursts. And intend to start using it while refactoring kni workload in controlplaneworker thread.

ol-imorozko:
Could you add this description to the commit message, please?

GeorgyKirichenko (changes requested):
I think we should get rid of limit on burst size using incorporated linked list inside `mbuf` meta.
In my opinion the Burst requires for three API methods:
 - build a list from `mbufs` array as DPDK returns a continuous array of them on read
 - an iterator through the list
 - a function to build continuous arrays of `mbufs` (with size limit) from the list

This approach allows us to inject new `mbufs` while pipeline processing.

Also, I think we need to have C-compatible API in order to support external processing modules. Despite the fact we do not have the ability now - we should keep this in mind.

TheRandomCharacter:
With moving to new pipeline this is no longer relevant

PR #176, opened 2024-04-16
Move management of kernel interfaces to separate class

TheRandomCharacter:


TheRandomCharacter:
This pull request wrongly places the responsibility of setting up queues onto the handle. Will rework as part of moving slow workload out of `cControlplane`

PR #177, opened 2024-04-16
clang-format changes: 120 columns limit, BlockIndent multiline args

TheRandomCharacter:


3Hren (changes requested):
What's the problem you're trying to solve with this PR? Does it affects formatting somewhere in the code?
I am strongly against reformatting the entire codebase without purpose, so the first thing we need to do is to is agree on a common code style.

TheRandomCharacter:
It was intended as a stick-it note for what out .clang-format is lacking. making a PR was a bad idea, this better suits an issue. Moreover adding clang-tidy will also require massive styling related changes and switching to refined .clang-format will be better done in one PR with clang-tidy changes. will add this changes to a discussion of clang-tidy

PR #190, opened 2024-05-06
Remove redundant memory_manager initialization

TheRandomCharacter:
For some reason trivial memory_manager initialization was done outside constructor. Always returning `eResult::success` initialization result was checked for really, truly, returning `eResult::success`.


ol-imorozko (approved):
I got a lil bit confused with the second sentence. I suggest to change it to smth like:

> Although always returning eResult::success, the initialization result was checked to ensure it really returned eResult::success.

Otherwise LGTM

GeorgyKirichenko (approved):

PR #194, opened 2024-05-07
Add two options to yanet-autotest-run to use gdb with control/dataplane

ol-imorozko:
This way, developers could use gdb without manually copying the command to run it

### DIFF
--- a/autotest/yanet-autotest-run.py
+++ b/autotest/yanet-autotest-run.py
@@ -27,15 +27,17 @@ class Autotest:
             for application in applications:
                 os.environ["PATH"] += f":{self.prefix}/{application}"
 
-    def wait_application(self, application):
-        for tries in range(1, 30):
+    def wait_application(self, application, is_gdb_enabled):
+        attempts = 0
+        while True:
             if os.system(f"yanet-cli version | grep --silent {application}") == 0:
                 return
             time.sleep(1)
-
-        print("ERROR: Failed to start application:", application)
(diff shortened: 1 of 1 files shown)


TheRandomCharacter:
This seems to have limited usefulness.
Maybe adding a script to start tmux session with panes containing dataplane, controlplane started under gdb and a test running pane  will be more flexible?

I'm currently using

```sh
#!/bin/bash
session="yanet-debug"
testid=$1
apt install --yes tmux
tmux new-session -s $session -d
tmux send-keys -t $session "gdb --args build_autotest/dataplane/yanet-dataplane -c autotest/units/001_one_port/dataplane.conf"
tmux split-window -h
tmux send-keys -t $session "gdb --args build_autotest/controlplane/yanet-controlplane -c autotest/units/001_one_port/$testid/controlplane.conf"
tmux split-window -v
tmux send-keys -t $session "build_autotest/autotest/yanet-autotest"
tmux attach -t $session

```

GeorgyKirichenko (approved):

ol-imorozko:
> Maybe adding a script to start tmux session with panes containing dataplane, controlplane started under gdb and a test running pane will be more flexible?

This script is certainly useful; however, I think it would be better to offer an option of starting gdb with yanet-autotest-run, because there will often be situations where you just need to quickly check how something works and not set up a fully-functional debug tmux session.

these changes won't conflict with such script though, we can add it in a separate PR

PR #197, opened 2024-05-20
Added counters for the route

stal76:
Now one can view the number and size of packets for each entry in the route.

	cli route counters
	cli route counters influxdb

Added a new autotest - 075_counters_rib, in the autotest.yaml one can see an example of statistics output.


autotest/units/001_one_port/075_counters_rib/gen.py
```diff
@@ -0,0 +1,42 @@
...
+	PcapWriter(filename)
+
+	for packets in packetsList:
+		if type(packets) == list:
+			for packet in packets:
+				packet.time = 0
+				wrpcap(filename, [p for p in packet], append=True)
+		else:
+			packets.time = 0
+			wrpcap(filename, [p for p in packets], append=True)
```
ol-imorozko, line 22: IMHO the following is better
```python
writer = PcapWriter(filename, append=True)
    
    for packets in packets_list:
        if isinstance(packets, list):
            for packet in packets:
                packet.time = 0
                writer.write(packet)
        else:
            packets.time = 0
            writer.write(packets)
            
    writer.close()
```

We will avoid reopening and closing the pcap file for every packet and we can use `isinstance` to check if an object is a list

cli/route.h
```diff
@@ -111,6 +112,51 @@ void get(const std::string& route_name,
 	table.print();
 }
 
+void counters(const std::optional<std::string>& format)
+{
+	bool use_table = true;
+	if (format != std::nullopt)
```
ol-imorozko, line 118: Use `format.has_value()`

cli/route.h
```diff
@@ -111,6 +112,51 @@ void get(const std::string& route_name,
...
+	{
+		if (*format == "influxdb")
+		{
+			use_table = false;
+		}
+		else if (*format != "table")
+		{
+			fprintf(stderr, "unknown output format: %s\n", format->c_str());
+			return;
```
ol-imorozko, line 127: Use
`std::cerr << "unknown output format: " << *format << std::endl;`

cli/route.h
```diff
@@ -111,6 +112,51 @@ void get(const std::string& route_name,
...
+		             "nexthop",
+		             "prefix",
+		             "counts",
+		             "size");
+
+		for (const auto& item : response)
+		{
+			const auto& [peer, nexthop, prefix, counts, size] = item;
+			table.insert(peer, nexthop, prefix, counts, size);
+		}
```
ol-imorozko, line 146: Just
```cpp
        for (const auto& [peer, nexthop, prefix, counts, size] : response)
        {
            table.insert(peer, nexthop, prefix, counts, size);
        }
```

cli/route.h
```diff
@@ -111,6 +112,51 @@ void get(const std::string& route_name,
...
+
+		table.print();
+	}
+	else
+	{
+		for (const auto& item : response)
+		{
+			const auto& [peer, nexthop, prefix, counts, size] = item;
```
ol-imorozko, line 154: Same thing here

ol-imorozko:
I would suggest splitting commit into two: one for the new functionality and the other for the test.

GeorgyKirichenko:
Overall the PR is ok but I would prefer to hold it until shared memory for counters and counter identifier allocator PR would be merged. Also I am not sure if available amount of counters is big enough so will test it on our production before.

Merging functionality and tests is Ok but please update the PR according to comments.

GeorgyKirichenko:
Also I would like to request telegraf output for the feature like telegraf dregress does

GeorgyKirichenko (approved):

PR #200, opened 2024-05-22
Nonterminating rules

ol-imorozko:


ol-imorozko:
I don't know why `041_firewall_keepstate_with_sync_multi_state_in_single_packet` is failing.
On my host in docker `./autotest/yanet-autotest-run.py --prefix=build_autotest autotest/units/001_one_port autotest/units/001_one_port/041_firewall_keepstate_with_sync_multi_state_in_single_packet/ -d` runs fine

common/actions.h
```diff
@@ -199,20 +200,26 @@ struct Action
 	}
 };
 
+namespace acl
```
GeorgyKirichenko, line 227: The idea is not clear to me so the question is why we need to have two kinds on action list instead of include check-state into the list.
Additional point here - we want to introduce more different actions (like dump, count, log, etc.) and I would like to process them all in a common way.

GeorgyKirichenko:
Could the PR be split into two parts:
 - keep-state -> recod-state + check-state
 - rule additional actions

common/actions.h
ol-imorozko: TLDR: that is to avoid extra paths in dataplane.
We process all actions in common way -- by calling the respective `execute` function. `check-state` is a unique one, cause it can return some return value, which we need to check. If we were to introduce that in the common handler function, this can drastically slow down the dataplane cause we will need to perform checks like `if (!retval)` 

When we create `Actions` objects before passing them to the dataplane, we know whether the check-state rule will be in the actions to be executed. Therefore, the `Actions` itself is a `std::variant` denoting whether we will execute check-state or not.

In either case we have to `check-state`, so this is done at the very beginning. If the check is successful, we get flow and execute the actions before check-state. This is implemented via a separate vector, **because that eliminates the need for unnecessary checks in dataplane like `if current_action != check-state action`**. If the check fails, we go to the standard package path.

autotest/units/001_one_port/036_firewall_keepstate/firewall.txt
```diff
@@ -2,6 +2,7 @@
 add skipto :IN ip from any to any in
 
 :IN
+add check-state
```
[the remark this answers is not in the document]
ol-imorozko: Done in the new pull request #241

TheRandomCharacter (dismissed):

GeorgyKirichenko (approved):

PR #204, opened 2024-06-03
Counters optimization

stal76:


controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+
+		void SetSize(uint16_t segment_size)
+		{
+			one_segment_size = segment_size;
+			total_segments = block_size / segment_size;
+			free_segments = total_segments;
+		}
+
+		uint16_t Allocate()
+		{
```
ol-imorozko, line 201: I don't like the idea of using some magic value as an error in this private method. We can use `std::optional`.
`std::optional<uint16_t> Allocate()`

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+			// calling the function implies that we are sure that there are free segments
+			if ((~group_mask) == 0)
+			{
+				return error_in_block_;
+			}
```
ol-imorozko, line 206:
```cpp
            if (group_mask.all())
            {
                return std::nullopt;
            }
```

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+
+			uint16_t group = __builtin_ctzll(~group_mask);
+			if ((group >= sizeof(masks) / sizeof(masks[0])) || ((~masks[group]) == 0))
+			{
+				return error_in_block_;
+			}
+			uint16_t index_in_group = __builtin_ctzll(~masks[group]);
+			uint16_t index = group * 64 + index_in_group;
```
ol-imorozko, line 214:
```cpp
            auto group = static_cast<uint16_t>(group_mask._Find_first());
            if (group >= masks.size() || masks[group].all())
            {
                return std::nullopt;
            }
            auto index_in_group = static_cast<uint16_t>(masks[group]._Find_first());
            uint16_t index = group * BITS_PER_GROUP + index_in_group;
```

`_Find_first` will use `__builtin_ctzll` so there will be no overhead, if that's a concern

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+
+			if (index >= total_segments)
+			{
+				return error_in_block_;
```
ol-imorozko, line 218: `return std::nullopt;`

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+			}
+
+			// set bits usage
+			ENABLE_BIT(masks[group], index_in_group);
+			if ((~masks[group]) == 0)
+			{
+				ENABLE_BIT(group_mask, group);
+			}
```
ol-imorozko, line 226:
```cpp
            masks[group].set(index_in_group);
            if (masks[group].all())
            {
                group_mask.set(group);
            }
```

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+
+			// change number of free segments
+			free_segments--;
+
+			return index;
+		}
+
+		uint16_t Free(uint16_t index)
```
ol-imorozko, line 234: Same thing with the return value here, but here we can just return `bool`

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+			DISABLE_BIT(masks[group], index);
+			if (masks[group] == 0)
+			{
+				DISABLE_BIT(group_mask, group);
+			}
+
+			// change number of free segments
+			free_segments++;
+
+			return 0;
```
ol-imorozko, line 259:
```cpp
            if (index >= total_segments)
            {
                return false;
            }

            uint16_t group = index >> GROUP_SHIFT;
            index &= GROUP_MASK;

            if (!masks[group].test(index))
            {
                return false;
            }

            // set bits usage
            masks[group].reset(index);
            if (masks[group].none())
            {
                group_mask.reset(group);
            }

            // change number of free segments
            free_segments++;
            return true;
```

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+		}
+	};
+
+	static constexpr uint32_t total_blocks_ = (IndexEnd - IndexBegin + BlockSize - 1) / BlockSize;
+	static constexpr uint32_t null_block_ = static_cast<uint32_t>(-1);
+	static constexpr uint16_t error_in_block_ = static_cast<uint16_t>(-1);
+
```
ol-imorozko, line 266: Don't need this

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+	OneBlock all_blocks_[total_blocks_];
+	OneSizeBlockInfo sizes_info_[MaxBufferSize + 1];
```
ol-imorozko, line 268:
```
    std::array<OneBlock, total_blocks_> all_blocks_;
    std::array<OneSizeBlockInfo, MaxBufferSize + 1> sizes_info_;
```

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+				// move block from list of free to blocks of this size
+				RemoveFromList(block_index, 0);
+				all_blocks_[block_index].SetSize(size);
+				InsertToList(block_index, size);
+			}
+		}
+
+		OneBlock& block = all_blocks_[size_info.head_block];
+		uint16_t index_in_block = block.Allocate();
+		if (index_in_block == error_in_block_)
```
ol-imorozko, line 84:
```cpp
        auto index_in_block = block.Allocate();
        if (!index_in_block.has_value())
```

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+		{
+			errors_internal_++;
+			return ErrorResult;
+		}
+		uint32_t result = size_info.head_block * BlockSize + index_in_block * size + IndexBegin;
```
ol-imorozko, line 89: `uint32_t result = size_info.head_block * BlockSize + index_in_block.value() * size + IndexBegin;`

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+		}
+		else if ((start - block_index * BlockSize - IndexBegin) % size != 0)
+		{
+			errors_external_++;
+			return false;
+		}
+
+		bool block_was_busy = (block.free_segments == 0);
+		uint16_t index = (start - block_index * BlockSize - IndexBegin) / size;
+		if (block.Free(index) == error_in_block_)
```
ol-imorozko, line 125: `if (!block.Free(index))`

ol-imorozko (changes requested):

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+
+	const OneSizeBlockInfo* GetBlocksStat() const
+	{
+		return sizes_info_;
+	}
+
+	std::pair<uint64_t, uint64_t> GetErrors() const
+	{
+		return {errors_external_, errors_internal_};
+	}
```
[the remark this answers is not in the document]
TheRandomCharacter: `[[nodiscard]]` is unnecessary here.
This function returns list of stored errors requested by caller, not errors encountered during execution of the function.

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
...
+#define ENABLE_BIT(v, b) (v) |= ((uint64_t)1) << (b)
+#define GET_BIT(v, b) (((v) >> (b)) & 1)
+#define DISABLE_BIT(v, b) (v) &= ~(((uint64_t)1) << (b))
+
+template<uint32_t IndexBegin, uint32_t IndexEnd, uint32_t BlockSize, uint32_t MaxBufferSize, uint32_t ErrorResult>
+class SegmentAllocator
+{
+public:
+	static_assert((BlockSize % 64) == 0);
+	static_assert((BlockSize <= 64 * 64) && (BlockSize >= 64));
```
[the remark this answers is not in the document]
TheRandomCharacter: `BITS_PER_GROUP` should be derived from mask size as `std::numeric_limits<decltype(OneBlock{}.group_mask)>::digits`

controlplane/unittest/segment_allocator.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+TEST(SegmentAllocator, BigSize)
+{
+	static constexpr uint32_t index_begin = 128;
+	static constexpr uint32_t index_end = 8 * 1024 * 1024;
+	using AllocatorBig = SegmentAllocator<index_begin, index_end, 64 * 64, 64, error_result>;
+
+	uint8_t data[sizeof(AllocatorBig)];
+	memset((void*)data, -1, sizeof(AllocatorBig));
+	AllocatorBig* allocator = new (data) AllocatorBig();
```
ol-imorozko, line 182: I think we need `alignas(AllocatorBig)` here so the memory is properly aligned for the type being constructed.

I mean `alignas(AllocatorBig) uint8_t data[sizeof(AllocatorBig)];`

Also there's no need to cast to `(void*)`, just use `memset(data, -1, sizeof(AllocatorBig));`. 
In fact, `alignas(AllocatorBig) uint8_t data[sizeof(AllocatorBig)]  = {-1};` is enough

ol-imorozko (changes requested):

controlplane/segment_allocator.h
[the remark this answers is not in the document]
ol-imorozko: - `OneBlock::group_mask` instead of `OneBlock{}.group_mask`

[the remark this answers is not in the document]
GeorgyKirichenko: The magic number are bad so please use `static constexpr` or corresponding defines inside the `config` file.

controlplane/segment_allocator.h
```diff
@@ -0,0 +1,308 @@
+#pragma once
+
+#include <cstdint>
+#include <cstring>
+
+#include <iostream>
+
+#define ENABLE_BIT(v, b) (v) |= ((uint64_t)1) << (b)
+#define GET_BIT(v, b) (((v) >> (b)) & 1)
+#define DISABLE_BIT(v, b) (v) &= ~(((uint64_t)1) << (b))
```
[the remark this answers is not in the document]
GeorgyKirichenko: Bitsets are great if corresponding bit count does not fit into standard unsigned types but if you will use only one predefined type C-way is acceptable for me.
However, the macroses above are not fully correct -there is no guaranties about `v` data type and b value, at least they should include uint64 in the macros names.
Also strong typed `static inline` functions are acceptable here.

controlplane/counter.h
```diff
@@ -1,108 +1,121 @@
...
-	{
-		static_assert(size_T <= YANET_CONFIG_COUNTER_FALLBACK_SIZE);
+		return index;
+	}
 
-		std::lock_guard<std::mutex> guard(counter_mutex);
+	bool Free(size_t start, size_t size)
+	{
+		auto iter_right = segments_start_.upper_bound({start, 0});
+		auto iter_left = (iter_right == segments_start_.begin() ? segments_start_.end() : std::prev(iter_right)); // если тот что справа самый первый, то левее уже никого
```
GeorgyKirichenko: Please use english for comments and place the comment just before the line

GeorgyKirichenko:
I like the new allocator but please find some notes bellow:
 - I feel you can squash the changes relating the allocator from the secon into the third commit
 - Please annotate classes and methods with comments
 - Small brief about allocator algorithm is required
 - The commit message headings could be more informational

controlplane/segment_allocator.h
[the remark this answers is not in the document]
TheRandomCharacter: `std::bitset` provides a well tested abstraction for "a set of N bits" with de facto standard interface. Not using it means that one is rolling ones own implementation. Is it tested as thoroughly? Is the interface intuiteve to use? And most importantly, does it provide additional or different functionality? If answer to any of this questions is no, on should prefer `std::bitset`

controlplane/segment_allocator.h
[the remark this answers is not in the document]
TheRandomCharacter: Placing this constants in `config` is questionable. This breaks `SegmentAllocatjr`  being independent. If configuring this values outside of `SegmentAllocator` is desirable exposing these variable as template parameters is a good option.

GeorgyKirichenko (approved):

Issue #35, opened 2023-10-03
Do not rebuild the whole real mapping ring on each update


GeorgyKirichenko:
`globalBase` maintains two arrays with service real mappings. For each instance corresponding mappings placed continuously without any space between two services. This means that any change in real state may lead to whole instance rebuild - and this is why there are two instances.

The main idea is to preallocate the maximal possible amount of space for each service. In the case any real update will touch only the real service partition of the array and do not require the whole mapping rebuild.
Furthermore, we could update mappings in place without maintaining two of mapping instances. The only thing we should care here is not to address invalid real for a service - so real identifiers for a service should be updated before the service mapping size.

GeorgyKirichenko:
This is prerequisite for #33

PR #205, opened 2024-06-05
Change balancer ring creation to block-based memory allocation

levon-a-akopian:
… memory for the ring is allocated in blocks


ol-imorozko:
Please use correct commit message conventions, that is, the first line should not be that long.
Smth like:
```
Change balancer ring creation to block-based memory allocation

This improves memory management efficiency by allocating ring memory in blocks.
```

README.md
```diff
@@ -33,6 +33,7 @@ docker pull yanetplatform/builder
 Add alias for run commands on docker:
 ```
 alias yanet-builder="docker run --rm -it -v /run/yanet:/run/yanet -v \$PWD:/project yanetplatform/builder"
+(for Mac: alias yanet-builder="docker run --platform linux/amd64 --rm -it -v /run/yanet:/run/yanet -v \$PWD:/project yanetplatform/builder")
```
ol-imorozko, line 36: This is not relevant change, please create another pull request

controlplane/balancer.cpp
```diff
@@ -668,14 +695,13 @@ void balancer_t::update_service(const balancer::generation_config_t& generation_
 
 			for (const auto& [real_ip, real_port, weight] : reals)
 			{
-				std::tuple<std::string, balancer::service_key_t, balancer::real_key_t> key = {module_name, {virtual_ip, proto, virtual_port}, {real_ip, real_port}};
+		    	balancer::real_key_global_t key = {module_name, {virtual_ip, proto, virtual_port}, {real_ip, real_port}};
```
ol-imorozko, line 698: `clang-format`

controlplane/balancer.h
```diff
@@ -26,9 +26,11 @@ using service_key_t = std::tuple<common::ip_address_t,
 using service_counter_key_t = std::tuple<std::string, ///< module_name
                                          service_key_t>;
 
-using real_counter_key_t = std::tuple<std::string, ///< module_name
-                                      service_key_t,
-                                      real_key_t>;
+using real_key_global_t = std::tuple<std::string, ///< module_name
+                                    service_key_t,
+                                    real_key_t>;
```
TheRandomCharacter, line 31: Is `tuple` really needed here? We should prefer `struct` to `tuple` in cases where we don't need to iterate over members.

If resorting to using tuple is unavoidable generous usage of type aliases makes code more readable:
```
using ModuleName = std::string
std::tuple<ModuleName>
```
is better than
```
std::tuple<std::string ///> module_name
>
```

README.md
GeorgyKirichenko: Yes, this is our overall politic to split non-relevant changes even if changes are small.

controlplane/balancer.cpp
```diff
@@ -354,6 +354,23 @@ void balancer_t::reload_after()
...
+			std::lock_guard<std::mutex> guard(reals_enabled_mutex);
+			// At the point real_updates and real_reload_updates
+			// may contain different sets of reals updates depending
+			// on update, flush and reload operation interposition.
+			// Besides the possibility that real_reload_updates is
+			// less than its counterpart it should be enough to
+			// flush only the first one and then reset both of them
+			// as real_reload_updates initial state already contains
+			// all changes was made just before the reload operation
+			// started. So we just want to rewu
```
GeorgyKirichenko, line 368: I feel the message is unfinished

GeorgyKirichenko (changes requested):
I feel the PR is mostly about #35 so please add corresponding line into the commit message like "Closed #35 or Relates #35"
Also please find some consideration about bellow:
 - Please move all irrelevant changes about incremental real update into separate PR as there is an #35
 -  Also #35 suggests that we could eliminate the second copy of real mapping ring as each service will have its reserved space inside the mapping but I could not see here any corresponding changes.

README.md
levon-a-akopian: Ok

GeorgyKirichenko (approved):

PR #208, opened 2024-06-13
timelimit for sendPakckets autotest stage. Packet receiving rewritten.

TheRandomCharacter:
New implementation bases on whole stage time limit (2s default equivalent of sum of waits used before) instead of hardcoded 1s header wait and 1s body wait.

Fixes logic of expecting read call on `SOCK_STREAM` to return whole header in one call and body on another single call.

Adds ability to override default 2s timelimit by adding a key-value pair `timelimit: <unsinged int seconds>` in sendPackets stage YAML.

### DIFF
--- a/autotest/autotest.cpp
+++ b/autotest/autotest.cpp
@@ -38,7 +38,8 @@ struct __attribute__((__packed__)) packHeader
 	uint32_t data_length;
 };
 
-using namespace nAutotest;
+namespace nAutotest
+{
 
 static const std::tuple<ip_address_t,
                         std::string,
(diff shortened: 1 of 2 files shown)


autotest/autotest.cpp
```diff
@@ -277,20 +277,43 @@ void tAutotest::sendThread(std::string interfaceName,
...
 
-static bool readPacket(int fd, pcap_pkthdr* header, u_char* data)
+bool readTimeLimited(int fd, void* data, ssize_t len, std::chrono::system_clock::time_point time_to_give_up)
 {
 	ssize_t ret = 0;
-	int i;
-
-	struct packHeader hdr;
-	for (i = 0; i < 1000; ++i)
+	char* buf = reinterpret_cast<char*>(data);
```
ol-imorozko, line 283: I don't like `void*`..
IMO it's better to ensure type safety with templated function and overloading:
```cpp
bool readTimeLimited(int fd, char* data, ssize_t len, std::chrono::system_clock::time_point time_to_give_up)
{
    ssize_t ret = 0;
    while (len > 0 && std::chrono::system_clock::now() < time_to_give_up)
    {
        ret = read(fd, data, len);
        switch (ret)
        {
            case 0:
                return false;
            case -1:
                if (errno == EAGAIN || errno == EWOULDBLOCK)
                {
                    std::this_thread::sleep_for(READ_WAIT_UNTIL_RETRY);
                }
                else
                {
                    return false;
                }
                break;
            default:
                len -= ret;
                data += ret;
                break;
        }
    }
    return len == 0;
}

template <typename T>
bool readTimeLimited(int fd, T& data, std::chrono::system_clock::time_point time_to_give_up)
{
    return readTimeLimited(fd, reinterpret_cast<char*>(&data), sizeof(T), time_to_give_up);
}
```

autotest/autotest.cpp
```diff
@@ -277,20 +277,43 @@ void tAutotest::sendThread(std::string interfaceName,
...
+
+	return len == 0;
+}
+
+static bool readPacket(int fd, pcap_pkthdr* header, u_char* data, Duration timelimit)
+{
+	auto time_to_give_up = std::chrono::system_clock::now() + timelimit;
+
+	struct packHeader hdr;
+	if (!readTimeLimited(fd, &hdr, sizeof(hdr), time_to_give_up))
```
ol-imorozko, line 316: `if (!readTimeLimited(fd, hdr, time_to_give_up))`

ol-imorozko (changes requested):

autotest/autotest.h
```diff
@@ -122,4 +129,12 @@ class tAutotest
 	uint64_t memorized_counter_value;
 };
 
+bool readTimeLimited(int fd, void* data, ssize_t len, std::chrono::system_clock::time_point time_to_give_up);
+
```
ol-imorozko, line 133: `char*` instead of `void*`

autotest/autotest.h
```diff
@@ -122,4 +129,12 @@ class tAutotest
...
+template<typename T>
+bool readTimeLimited(int fd, T& data, std::chrono::system_clock::time_point time_to_give_up)
+{
+	return readTimeLimited(fd, reinterpret_cast<void*>(&data), sizeof(data), time_to_give_up);
```
ol-imorozko, line 137: `char*` instead of `void*`

ol-imorozko (changes requested):

GeorgyKirichenko (approved):

PR #209, opened 2024-06-24
Cleanup

ol-imorozko:


ol-imorozko:
Strange, `clang-format` for me does not show any errors.
I've used `find . -type f \( -name "*.h" -o -name "*.cpp" \) -not -path "./build_autotest/*" -exec clang-format-18 -i {} +`

GeorgyKirichenko (approved):
LGTM

PR #210, opened 2024-06-25
Added support for RFC 5549, the ability to encapsulate IPv4 traffic in IPv6.

stal76:
Added support for RFC 5549, the ability to encapsulate IPv4 traffic in IPv6.

- Added encapsulation capability for "route" and "route tunnel".
- Added 2 autotests.


autotest/units/001_one_port/076_rfc_5549_route/gen.py
```diff
@@ -0,0 +1,35 @@
...
+	PcapWriter(filename)
+
+	for packets in packetsList:
+		if type(packets) == list:
+			for packet in packets:
+				packet.time = 0
+				wrpcap(filename, [p for p in packet], append=True)
+		else:
+			packets.time = 0
+			wrpcap(filename, [p for p in packets], append=True)
```
ol-imorozko, line 22: It's better to use PcapWriter's method `write` instead of opening file each time:
https://github.com/yanet-platform/yanet/pull/197#discussion_r1608293514

GeorgyKirichenko (approved):

PR #211, opened 2024-06-26
wlc moved from dataplane to controlplane

levon-a-akopian:


ol-imorozko:
Could you please improve the commit message?

Change `wlc moved to from dataplane to controlplane` to `Move wlc from dataplane to controlplane`

Also, add a short description explaining why this change is needed, please

controlplane/balancer.cpp
```diff
@@ -886,6 +886,12 @@ void balancer_t::flush_reals(common::idp::updateGlobalBaseBalancer::request& bal
 					}
 				}
 
+				auto reals_wlc = reals_wlc_weight.find(key);
+				if (reals_wlc != reals_wlc_weight.end() && reals_wlc->second.has_value())
+				{
+					effective_weight = reals_wlc->second.value();
+				}
```
ol-imorozko, line 893: We will have code duplication here with the following lines:
```cpp
auto it = reals_enabled.find(key);
if (it != reals_enabled.end())
{
	if (it->second.has_value())
	{
		effective_weight = it->second.value();
	}
}
```

I think we can add a private method in `balancer_t` like
```cpp
 template <typename Map>
    std::optional<typename Map::mapped_type::value_type> get_effective_weight(const Map& map, const balancer::real_key_global_t& key) const {
        auto it = map.find(key);
        if (it != map.end() && it->second.has_value()) {
            return it->second.value();
        }
        return std::nullopt;
    }
```

And then use it, for example, with compound initialization:
```cpp
if (auto found_weight = get_effective_weight(reals_wlc_weight, key); found_weight) {
    effective_weight = *found_weight;
}
```

controlplane/balancer.cpp
```diff
@@ -929,8 +935,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+				{
+					auto it = reals_enabled.find(key);
+					if (it != reals_enabled.end())
+					{
+						if (it->second.has_value())
+						{
+							effective_weight = it->second.value();
+						}
+					}
+				}
```
ol-imorozko, line 1005:
```cpp
auto found_weight = get_effective_weight(reals_enabled, key);
uint32_t effective_weight = found_weight ? *found_weight : weight;
```

controlplane/balancer.cpp
```diff
@@ -929,8 +935,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+				for (auto& [socket_id, real_connections] : balancer_real_connections)
+				{
+					(void)socket_id;
+
+					auto it = real_connections.find(real_connections_key);
+					if (it == real_connections.end())
+					{
+						continue;
+					}
+					connections += it->second;
```
ol-imorozko, line 1031: IMHO
```cpp
if (it != real_connections.end())
{
    connections += it->second;
}
```
is cleaner, but feel free no ignore

ol-imorozko (changes requested):

controlplane/balancer.cpp
```diff
@@ -926,8 +928,143 @@ void balancer_t::reconfigure_wlc_thread()
...
+		                  flags,
+		                  ipv4_outer_source_network,
+		                  ipv6_outer_source_network,
+		                  reals] : balancer.services)
+		{
+			(void)flags;
+			(void)version;
+			(void)forwarding_method;
+			(void)ipv4_outer_source_network;
+			(void)ipv6_outer_source_network;
```
ol-imorozko, line 969: use `YANET_GCC_BUG_UNUSED`

controlplane/balancer.cpp
```diff
@@ -926,8 +928,143 @@ void balancer_t::reconfigure_wlc_thread()
...
+				common::idp::balancer_real_connections::real_key_t real_connections_key = {balancer.balancer_id,
+				                                                                           virtual_ip,
+				                                                                           proto,
+				                                                                           virtual_port.value(),
+				                                                                           real_ip,
+				                                                                           real_port.value()};
+				uint32_t connections = 0;
+				for (auto& [socket_id, real_connections] : balancer_real_connections)
+				{
+					(void)socket_id;
```
ol-imorozko, line 1009: and here

ol-imorozko (dismissed):

controlplane/balancer.h
```diff
@@ -133,7 +135,23 @@ class balancer_t : public module_t, common::icp_proto::BalancerService
...
+	uint32_t calculate_wlc_weight(uint32_t weight, uint32_t connections, uint32_t weight_sum, uint32_t connection_sum, uint32_t wlc_power);
+
+	template<typename Map>
+	std::optional<typename Map::mapped_type::value_type> get_effective_weight(const Map& map, const balancer::real_key_global_t& key) const
+	{
+		auto it = map.find(key);
+		if (it != map.end() && it->second.has_value())
+		{
+			return it->second.value();
+		}
```
ol-imorozko, line 154: isn't that equivalent to just 
```suggestion
		if (it != map.end())
		{
			return it->second;
		}
```

ol-imorozko (approved):

GeorgyKirichenko (approved):
Please do not make  refactorings like `YANET_GCC_BUG_UNUSED` renaming in the same PR where functional changes are done as this complicates the review process.

Or if you feel such refactoring essential then each commit should be atomic and enabled to review separately , without any further `according to review` references in the following commits.

PR #217, opened 2024-07-02
Multiple slow workers

TheRandomCharacter:


common/type.h
```diff
@@ -21,6 +21,7 @@
 using tCoreId = uint32_t;
 using tSocketId = uint32_t;
 using tPortId = uint16_t;
+constexpr tPortId INVALID_PORT_ID = std::numeric_limits<tPortId>::max();
```
ol-imorozko, line 24: How are we using `INVALID_PORT_ID`? We do not check if the port is valid in any methods. The only usage is marking `m_kni_port` as "Invalid" in the Remove method and then checking its validity in the destructor. But we can achieve the same with a boolean flag and avoid using the maximum u16 value as a marker.

Or is it because some methods __could__ be called on an object with an invalid port, and we want DPDK functions to 100% result in an error? However, if all the code is correct, we shouldn't call such methods on an already invalid object, right?

dataplane/kernel_interface_handle.h
```diff
@@ -0,0 +1,70 @@
...
+
+	KernelInterfaceHandle() noexcept = default;
+
+public:
+	~KernelInterfaceHandle();
+	KernelInterfaceHandle(const KernelInterfaceHandle&) = delete;
+	KernelInterfaceHandle(KernelInterfaceHandle&& other) noexcept;
+	[[nodiscard]] static std::optional<KernelInterfaceHandle>
+	MakeKernelInterfaceHandle(
+	        const std::string& name,
```
ol-imorozko, line 25: `std::string_view`

dataplane/kernel_interface_handle.cpp
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return true;
+}
+
+KernelInterfaceHandle::~KernelInterfaceHandle()
+{
+	if (Valid())
+		Remove();
+}
+KernelInterfaceHandle::KernelInterfaceHandle(KernelInterfaceHandle&& other)
```
ol-imorozko, line 38: Separate definitions with an empty line, please.
Sometimes it is done, sometimes not, as seen here

ol-imorozko (changes requested):
First commit:

dataplane/kernel_interface_handle.h
```diff
@@ -0,0 +1,70 @@
...
+	static rte_eth_conf DefaultConfig()
+	{
+		rte_eth_conf eth_conf;
+		memset(&eth_conf, 0, sizeof(eth_conf));
+		return eth_conf;
+	}
+	bool Configure(const rte_eth_conf& eth_conf) noexcept;
+	bool CloneMTU(const uint16_t) const;
+	void MarkInvalid() { m_kni_port = INVALID_PORT_ID; }
+	[[nodiscard]] bool Valid() { return m_kni_port != INVALID_PORT_ID; }
```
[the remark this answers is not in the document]
ol-imorozko: * this is about `[[nodiscard]] bool Valid() { return m_kni_port != INVALID_PORT_ID; }`

dataplane/dataplane.cpp
```diff
@@ -1024,6 +1165,26 @@ eResult cDataPlane::initQueues()
...
+	for (auto& it : kni_interface_handles)
+	{
+		uint16_t port_id = it.first;
+		for (std::size_t i = 0, max = 1; i < max; ++i)
+		{
+			const auto& socket_id = rte_eth_dev_socket_id(port_id);
+			KNIAddTxQueue(i, socket_id);
+		}
+
+		for (std::size_t i = 0, max = 1; i < max; ++i)
```
ol-imorozko: `max = 1` seems strange. I think we should define some constexpr value and name it accordingly

ol-imorozko (changes requested):
Third commit: change "refactored" in the title to something more meaningful, please. Like "Move Kernel Network Interfaces initialization to cDataPlane"

common/type.h
```diff
@@ -2453,6 +2453,19 @@ struct stats_t
...
+		current_count_packets += other.current_count_packets;
+		total_overflow_packets += other.total_overflow_packets;
+		not_fragment_packets += other.not_fragment_packets;
+		empty_packets += other.empty_packets;
+		flow_overflow_packets += other.flow_overflow_packets;
+		intersect_packets += other.intersect_packets;
+		unknown_network_type_packets += other.unknown_network_type_packets;
+		timeout_packets += other.timeout_packets;
+		return *this;
+	}
```
ol-imorozko, line 2468: I suggest moving this change into a separate commit. "Fragmentation untied from control and dataplane" does not seems like an appropriate place.

common/type.h
```diff
@@ -2504,6 +2517,25 @@ struct stats_t
 	uint64_t tcp_ok;
 	uint64_t tcp_timeout_sessions;
 	uint64_t tcp_unknown_sessions;
+
+	stats_t& operator+=(const stats_t& other)
+	{
+		bad_decap_transport += other.bad_decap_transport;
```
ol-imorozko, line 2523: same here

dataplane/fragmentation.h
```diff
@@ -37,30 +38,38 @@ using value_t = std::tuple<std::map<uint32_t, ///< range_from
...
-class fragmentation_t
+class Fragmentation
 {
 public:
-	fragmentation_t(cControlPlane* controlPlane, cDataPlane* dataPlane);
-	~fragmentation_t();
+	using OnCollected = std::function<void(rte_mbuf*, const common::globalBase::tFlow&)>;
+	Fragmentation(OnCollected callback);
+	Fragmentation(OnCollected callback, const FragmentationConfig& cfg);
+	Fragmentation(Fragmentation&& other);
```
ol-imorozko, line 47: `noexcept`

dataplane/fragmentation.h
```diff
@@ -37,30 +38,38 @@ using value_t = std::tuple<std::map<uint32_t, ///< range_from
...
+	~Fragmentation();
+
+	Fragmentation& operator=(Fragmentation&& other);
```
ol-imorozko, line 50: `noexcept`

dataplane/fragmentation.h
```diff
@@ -37,30 +38,38 @@ using value_t = std::tuple<std::map<uint32_t, ///< range_from
...
 	void insert(rte_mbuf* mbuf);
 	void handle();
 
 protected:
-	bool isTimeout(const fragmentation::value_t& value);
-	bool isCollected(const fragmentation::value_t& value);
-	bool isIntersect(const fragmentation::value_t& value, const uint32_t& range_from, const uint32_t& range_to);
+	bool isTimeout(const value_t& value);
+	bool isCollected(const value_t& value);
+	bool isIntersect(const value_t& value, const uint32_t& range_from, const uint32_t& range_to);
```
ol-imorozko, line 63: `const`

dataplane/fragmentation.h
```diff
@@ -37,30 +38,38 @@ using value_t = std::tuple<std::map<uint32_t, ///< range_from
...
-	common::fragmentation::stats_t getStats();
+	common::fragmentation::stats_t getStats() const;
+	OnCollected& Callback() { return callback_; }
+	void Configure(const FragmentationConfig& cfg) { config_ = cfg; }
 
 	void insert(rte_mbuf* mbuf);
 	void handle();
 
 protected:
-	bool isTimeout(const fragmentation::value_t& value);
```
ol-imorozko, line 55: Unrelated, but we're making this class better in this commit anyway, so I guess it's fine:
Let's change `protected` to `private` here. We're not using this class in inheritance in the project

dataplane/fragmentation.cpp
```diff
@@ -1,19 +1,39 @@
...
-                                 cDataPlane* dataPlane) :
-        controlPlane(controlPlane),
-        dataPlane(dataPlane)
+namespace fragmentation
 {
-	memset(&stats, 0, sizeof(stats));
+
+Fragmentation::Fragmentation(OnCollected callback) :
+        callback_{callback}
+{}
```
ol-imorozko, line 11: maybe memset `stats_` here too?

dataplane/fragmentation.cpp
```diff
@@ -1,19 +1,39 @@
...
-fragmentation_t::fragmentation_t(cControlPlane* controlPlane,
-                                 cDataPlane* dataPlane) :
-        controlPlane(controlPlane),
-        dataPlane(dataPlane)
+namespace fragmentation
 {
-	memset(&stats, 0, sizeof(stats));
+
+Fragmentation::Fragmentation(OnCollected callback) :
+        callback_{callback}
```
ol-imorozko, line 10: `callback_(std::move(callback))`

dataplane/fragmentation.cpp
```diff
@@ -1,19 +1,39 @@
...
+{}
+
+Fragmentation::Fragmentation(OnCollected callback, const FragmentationConfig& cfg) :
+        callback_{callback},
```
ol-imorozko, line 14: `callback_(std::move(callback))`

dataplane/fragmentation.cpp
```diff
@@ -1,19 +1,39 @@
...
+        config_{cfg}
+{
+	memset(&stats_, 0, sizeof(stats_));
+}
+
+Fragmentation::Fragmentation(Fragmentation&& other)
+{
+	*this = std::move(other);
+}
```
ol-imorozko, line 23: Same problem as in the first commit, it's better to use
```cpp
Fragmentation::Fragmentation(Fragmentation&& other) noexcept
    : callback_(std::move(other.callback_)),
      config_(std::move(other.config_)),
      stats_(std::move(other.stats_)),
      fragments_(std::move(other.fragments_)) {}
  ```

File: dataplane/fragmentation.cpp [hunk 16]
```diff
@@ -1,19 +1,39 @@
...
+
+Fragmentation& Fragmentation::operator=(Fragmentation&& other)
+{
+	std::swap(callback_, other.callback_);
+	config_ = other.config_;
+	stats_ = other.stats_;
+	std::swap(fragments_, other.fragments_);
+	return *this;
 }
```
ol-imorozko, line 32:
```cpp
Fragmentation& Fragmentation::operator=(Fragmentation&& other) noexcept {
    if (this != &other) {
        callback_ = std::move(other.callback_);
        config_ = std::move(other.config_);
        stats_ = std::move(other.stats_);
        fragments_ = std::move(other.fragments_);
    }
    return *this;
}
```

ol-imorozko (changes requested):
Fifth commit:
Use present tense in commit titles. Smth like:
```
Untie fragmentation from control and data planes.

Also, apply Google coding style to private members by adding a `_` suffix.
```

dataplane/fragmentation.cpp
ol-imorozko: Actually we can just use value initialization from `stats_` field: `common::fragmentation::stats_t stats_{};`
`stats_t`does not have any default arguments and will zero itself

dataplane/fragmentation.cpp
```diff
@@ -1,19 +1,39 @@
 #include "fragmentation.h"
 #include "common.h"
 #include "controlplane.h"
```
ol-imorozko, line 3: We got rid of controlplane, so now this include became unused. Delete it, please

common/ctree.h
```diff
@@ -323,6 +354,37 @@ class ctree
...
+			}
+
+			for (const auto& [key, value] : reinterpret_cast<const node_t*>(other)->convert)
+			{
+				if (convert.find(key) == convert.end())
+				{
+					convert.emplace(key, value);
+				}
+			}
+		}
```
ol-imorozko, line 386: I think we can just use `static_cast` like in the rest of the code. I don't like `reinterpret cast`.
Smth like:
```cpp
const auto* other_node = static_cast<const node_t*>(other);
//...
for (const auto& [key, node] : other_node->next)
{
    //..
		next.emplace(key, new actual(*static_cast<actual*>(node)));
//..
for (const auto& [key, value] : other_node->convert)
//...
```

common/ctree.h
```diff
@@ -336,4 +398,13 @@ class ctree
...
 };
 
+inline typename std::array<uint64_t, 4>& operator+=(std::array<uint64_t, 4>& a, const std::array<uint64_t, 4>& b)
+{
+	for (std::size_t i = 0, e = a.size(); i < e; ++i)
+	{
+		a[i] += b[i];
+	}
+	return a;
+}
```
ol-imorozko, line 408: Why four specifically?
Can't we just do smth like
```cpp
template <std::size_t N>
inline std::array<uint64_t, N>& operator+=(std::array<uint64_t, N>& a, const std::array<uint64_t, N>& b)
{
    for (std::size_t i = 0; i < N; ++i)
    {
        a[i] += b[i];
    }
    return a;
}
```
?

ol-imorozko (changes requested):
Sixth commit:
Also change commit message a bit, it should be in present tense:

> Add merge and move operations to ctree
> 
> Also introduce an alias for std::array<uint64_t, values_size_T> and
> use a custom += operator for an arrays.

ol-imorozko (changes requested):
Seventh commit, improve commit message slightly:

> Move hashtable stats out of class
> 
> Declare stats as a separate type to prevent interference with nlohmann's ability to find conversion functions.

dataplane/dregress.cpp
```diff
@@ -4,19 +4,45 @@
...
+
+	other.connections = nullptr;
+}
 
-dregress_t::dregress_t(cControlPlane* controlplane,
-                       cDataPlane* dataplane) :
-        controlplane(controlplane),
-        dataplane(dataplane)
+dregress_t& dregress_t::operator=(dregress_t&& other)
 {
```
ol-imorozko, line 28: `noexcept`

dataplane/dregress.h
```diff
@@ -43,44 +43,76 @@ using direction_t = std::tuple<common::ip_prefix_t, ///< prefix
...
 
 class dregress_t
 {
 public:
-	dregress_t(cControlPlane* controlplane, cDataPlane* dataplane);
+	dregress_t(dataplane::SlowWorker* slow, cDataPlane* dataplane, uint32_t gc_step);
+	dregress_t(dregress_t&& other);
 	~dregress_t();
 
+	dregress_t& operator=(dregress_t&& other);
```
ol-imorozko, line 81: `noexcept`

dataplane/dregress.h
```diff
@@ -43,44 +43,76 @@ using direction_t = std::tuple<common::ip_prefix_t, ///< prefix
...
 	bool tcp_parse(rte_mbuf* mbuf, uint16_t& rtt, uint32_t& loss_count, uint32_t& ack_count);
-
-	common::idp::get_dregress_counters::response get_dregress_counters();
-	void limits(common::idp::limits::response& response);
+	dregress::LimitsStats limits() const;
+	const common::dregress::stats_t& Stats() const { return stats; }
+	const dregress::ConnTable* Connections() const { return connections; }
+	[[nodiscard]] std::lock_guard<std::mutex> LockCounters() { return std::lock_guard{counters_mutex}; }
+	const common::dregress::counters_t& Counters4() { return counters_v4; }
+	const common::dregress::counters_t& Counters6() { return counters_v4; }
```
ol-imorozko, line 93: `const`

dataplane/dregress.cpp
```diff
@@ -4,19 +4,45 @@
...
 #include "dregress.h"
 #include "metadata.h"
-#include "worker.h"
+#include "slow_worker.h"
+
+dregress_t::dregress_t(dataplane::SlowWorker* slow, cDataPlane* dataplane, uint32_t gc_step) :
+        slow_worker_{slow},
+        dataplane{dataplane},
+        stats{},
+        connections{new dregress::ConnTable},
```
ol-imorozko, line 16: That's bad. I suggest avoiding raw pointers alongside allocating memory with `new` in constructor and calling `delete` in destructor. That's what `std::unique_ptr` does.
Here it will be `connections_(std::make_unique<dregress::ConnTable>())`

dataplane/dregress.cpp
```diff
@@ -4,19 +4,45 @@
...
-        controlplane(controlplane),
-        dataplane(dataplane)
+dregress_t& dregress_t::operator=(dregress_t&& other)
 {
-	memset(&stats, 0, sizeof(stats));
-	connections = new dataplane::hashtable_chain_spinlock_t<dregress::connection_key_t, dregress::connection_value_t, YANET_CONFIG_DREGRESS_HT_SIZE, YANET_CONFIG_DREGRESS_HT_EXTENDED_SIZE, 4, 4>();
+	slow_worker_ = other.slow_worker_;
+	dataplane = other.dataplane;
+	stats = other.stats;
+	std::swap(connections, other.connections);
```
ol-imorozko, line 32: when `connections` will be `std::unique_ptr` instead of raw pointer, use `std::move` here

dataplane/dregress.h
```diff
@@ -43,44 +43,76 @@ using direction_t = std::tuple<common::ip_prefix_t, ///< prefix
...
 	void handle();
 
 	std::optional<dregress::direction_t> lookup(rte_mbuf* mbuf);
 	bool tcp_parse(rte_mbuf* mbuf, uint16_t& rtt, uint32_t& loss_count, uint32_t& ack_count);
-
-	common::idp::get_dregress_counters::response get_dregress_counters();
-	void limits(common::idp::limits::response& response);
+	dregress::LimitsStats limits() const;
+	const common::dregress::stats_t& Stats() const { return stats; }
+	const dregress::ConnTable* Connections() const { return connections; }
```
ol-imorozko, line 90: `return connections_.get();`

dataplane/dregress.h
```diff
@@ -43,44 +43,76 @@ using direction_t = std::tuple<common::ip_prefix_t, ///< prefix
...
 	constexpr static double median_multiplier = 0.01;
 
 	common::dregress::stats_t stats;
-	dataplane::hashtable_chain_spinlock_t<dregress::connection_key_t,
-	                                      dregress::connection_value_t,
-	                                      YANET_CONFIG_DREGRESS_HT_SIZE,
-	                                      YANET_CONFIG_DREGRESS_HT_EXTENDED_SIZE,
-	                                      4,
-	                                      4>* connections;
+	dregress::ConnTable* connections;
```
ol-imorozko, line 107: `std::unique_ptr<dregress::ConnTable> connections_;`

dataplane/slow_worker.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "common/type.h"
+#include "controlplane.h"
+#include "worker.h"
+
+namespace dataplane
+{
+
+SlowWorker::SlowWorker(cControlPlane* cplane) :
+        cplane_{cplane} {}
```
ol-imorozko: `explicit` to avoid accidental implicit casts

ol-imorozko (changes requested):
Ninth commit:
improve commit message, smth like

> Refactor dregress to depend on slow worker

> The SlowWorker abstraction now handles interactions with the slow worker.
> Updated cControlPlane to initialize and use the SlowWorker instance.

dataplane/utils.h
```diff
@@ -0,0 +1,48 @@
...
+	It operator->() { return curr_; }
+	RoundRobinIterator& operator++()
+	{
+		if (++curr_ == end_)
+		{
+			curr_ = begin_;
+		}
+		return *this;
+	}
+};
```
ol-imorozko, line 46: How is this class used? I mean, we're only using it's `->` operator and not an increment one.
Is it for some future convenience?

dataplane/worker_gc.h
```diff
@@ -6,19 +6,23 @@
...
 #include "common/idp.h"
+#include "common/static_vector.h"
 #include "hashtable.h"
 
 class worker_gc_t
 {
 public:
-	worker_gc_t(cDataPlane* dataplane);
+	using PortToSocketArray = std::array<tSocketId, CONFIG_YADECAP_PORTS_SIZE>;
+	using SamplersVector = utils::StaticVector<samples::Sampler*, YANET_CONFIG_MAX_SAMPLED_WORKERS_PER_GC>;
```
ol-imorozko, line 20: Missing `#include "samples.h"`, we should add it

dataplane/worker_gc.h
```diff
@@ -6,19 +6,23 @@
...
+	worker_gc_t(const ConfigValues& cfg, const PortToSocketArray& pts, SamplersVector&& samplers);
```
ol-imorozko, line 21: same with `config_values.h`

dataplane/worker_gc.h
```diff
@@ -44,14 +48,16 @@ class worker_gc_t
...
-	void send_to_slowworker(rte_mbuf* mbuf, const common::globalBase::eFlowType& flow_type);
-	void send_to_slowworker(rte_mbuf* mbuf, const common::globalBase::tFlow& flow);
+	void SendToSlowWorker(rte_mbuf* mbuf);
 
 	void handle_samples();
 
 public:
-	cDataPlane* dataplane;
-	cControlPlane* controlplane;
+	[[nodiscard]] std::optional<dpdk::RingConn<rte_mbuf*>> RegisterSlowWorker(const std::string& name,
```
ol-imorozko, line 56: `std::string_view`

dataplane/worker_gc.cpp
```diff
@@ -83,37 +73,59 @@ eResult worker_gc_t::init(const tCoreId& core_id,
...
 
-	ring_to_slowworker = rte_ring_create(("r_tsw_" + std::to_string(core_id)).c_str(),
-	                                     dataplane->getConfigValues().ring_normalPriority_size,
-	                                     socket_id,
-	                                     RING_F_SP_ENQ | RING_F_SC_DEQ);
-	if (!ring_to_slowworker)
+	return eResult::success;
+}
+
+[[nodiscard]] std::optional<dpdk::RingConn<rte_mbuf*>> worker_gc_t::RegisterSlowWorker(const std::string& name,
```
ol-imorozko, line 79: `std::string_view`

common/static_vector.h
```diff
@@ -0,0 +1,139 @@
+#pragma once
+#include <array>
+
+#include "utils.h"
```
ol-imorozko, line 4: `utils.h`? What for?
Also don't need `array`, just `#include <algorithm>` is enough.

dataplane/worker_gc.cpp
```diff
@@ -10,18 +10,19 @@
...
         socket_id(-1),
         iteration(0),
         current_base_id(0),
         local_base_id(0),
-        ring_to_slowworker(nullptr),
-        ring_to_free_mbuf(nullptr),
-        callback_id(0)
+        toSlowWorker_(toSlowWorkers_.begin(), toSlowWorkers_.end()),
+        port_id_to_socket_id{pts},
+        samplers_{samplers},
```
ol-imorozko, line 22: `samplers` is an lvalue in this expression, we need `std::move` there

dataplane/dpdk.h
```diff
@@ -0,0 +1,82 @@
...
+#include <string>
+
+#include <rte_ring.h>
+
+#include "common/type.h"
+
+namespace dpdk
+{
+template<typename T>
+class Ring
```
ol-imorozko, line 12: Why is this a template class? We're not using any specialization except for `rte_mbuf*` anyway

dataplane/worker_gc.h
```diff
@@ -6,19 +6,23 @@
...
 public:
-	worker_gc_t(cDataPlane* dataplane);
+	using PortToSocketArray = std::array<tSocketId, CONFIG_YADECAP_PORTS_SIZE>;
+	using SamplersVector = utils::StaticVector<samples::Sampler*, YANET_CONFIG_MAX_SAMPLED_WORKERS_PER_GC>;
+	worker_gc_t(const ConfigValues& cfg, const PortToSocketArray& pts, SamplersVector&& samplers);
 	~worker_gc_t();
 
 	eResult init(const tCoreId& core_id, const tSocketId& socket_id, const dataplane::base::permanently& base_permanently, const dataplane::base::generation& base);
-	void start();
+	void start(pthread_barrier_t* runBarrier);
```
ol-imorozko: snake_case for method arguments

ol-imorozko (changes requested):
fifteenth commit:
Also please put verb first as in your previous commits: `Refactor GC for per-NUMA slow workers` instead of `GC refactor for pernuma slow workers`

dataplane/dataplane.cpp
```diff
@@ -1868,6 +1875,124 @@ eResult cDataPlane::parseJsonPorts(const nlohmann::json& json)
...
+	{
+		for (auto j : cpw.value())
+		{
+			if (!add_worker(j))
+			{
+				return std::nullopt;
+			}
+		}
+	}
+	return std::optional{std::move(result)};
```
ol-imorozko, line 2067: This can violate copy elision, better to just use `return result`;

dataplane/dataplane.cpp
```diff
@@ -1868,6 +1875,124 @@ eResult cDataPlane::parseJsonPorts(const nlohmann::json& json)
...
+std::set<InterfaceName> cDataPlane::workerInterfacesToService()
+{
+	std::set<InterfaceName> res;
+	for (const auto& p : config.workers)
+	{
+		const auto& ifaces = p.second;
+		for (const auto& i : ifaces)
+		{
+			res.emplace(i);
+		}
```
ol-imorozko, line 2088:
```cpp
    for (const auto& p : config.workers)
    {
        res.insert(p.second.begin(), p.second.end());
    }
```

dataplane/dataplane.cpp
```diff
@@ -1868,6 +1875,124 @@ eResult cDataPlane::parseJsonPorts(const nlohmann::json& json)
...
+			YADECAP_LOG_ERROR("controlPlaneWorker entry \"interfaces\" contains invalid type.");
+			return std::nullopt;
+		}
+		worker_ports.insert(InterfaceName{j});
+	}
+	return std::optional{
+	        std::pair<tCoreId, std::set<InterfaceName>>{
+	                core,
+	                std::move(worker_ports)}};
+}
```
ol-imorozko: IMHO just `return std::make_pair(core, std::move(worker_ports);` is better

dataplane/dataplane.cpp
```diff
@@ -1969,9 +2094,55 @@ eResult cDataPlane::checkConfig()
...
+bool cDataPlane::checkControlPlaneWorkersConfig()
+{
+	std::set<InterfaceName> assigned;
+	std::set<InterfaceName> to_assign = workerInterfacesToService();
+	bool result = true;
+	for (const auto& [core, worker_ports] : config.controlplane_workers)
+	{
+		for (const auto& p : worker_ports)
+		{
+
```
ol-imorozko, line 2305: remove extra empty line

dataplane/dataplane.cpp
```diff
@@ -1969,9 +2094,55 @@ eResult cDataPlane::checkConfig()
...
+			if (to_assign.find(p) == to_assign.end())
+			{
+				YADECAP_LOG_ERROR("Control plane worker config contains port not assigned to fast worker: core: %d, port: %s\n", core, p.c_str());
+				result = false;
+				continue;
+			}
+
+			assigned.emplace(p);
+			to_assign.erase(p);
```
ol-imorozko, line 2321: Instead of checking with `find` and then doing `emplace` and `erase`, we can just use return value of those methods:
```cpp
            if (!assigned.insert(p).second)
            {
                YADECAP_LOG_ERROR("Duplicate port in control plane worker config: core: %d, port: %s\n", core, p.c_str());
                result = false;
                continue;
            }

            if (to_assign.erase(p) == 0)
            {
                YADECAP_LOG_ERROR("Control plane worker config contains port not assigned to fast worker: core: %d, port: %s\n", core, p.c_str());
                result = false;
                continue;
            }
```

`insert` returns a pair containing an iterator to the inserted element and a flag indicating whether the insertion took place.
`erase` returns the number of elements removed, which is 0 if the element was not found.

ol-imorozko (changes requested):
Sixteenth commit:
Don't forget to remove #180, as this change is incorporated here.
Also add a verb to commit title please: `Control plane workers config parsing` -> `Implement control plane workers config parsing`

ol-imorozko (changes requested):
Nineteenth commit:
Commit message:

> Refactor worker RX ports
> 
> - Change unnamed embedded struct describing RX worker port to Endpoint
> - Change container of RX worker ports to StaticVector

ol-imorozko (changes requested):
Twentieth commit:
Correct your commit message, please:

> Move RX queue initialization out of workers
> 
> Configuring RX queues at the dataplane level allows us to get rid of
> synchronization barriers and simplify the startup logic.

dataplane/controlplane.cpp
```diff
@@ -291,17 +270,17 @@ common::idp::getWorkerStats::response cControlPlane::getWorkerStats(const common
...
 			{
 				portsStats[portIter.first] = worker->statsPorts[portIter.first];
 			}
 
-			response[coreId] = {worker->iteration,
-			                    worker->stats,
-			                    portsStats};
+			response[worker->coreId] = {worker->iteration,
+			                            worker->stats,
+			                            portsStats};
```
ol-imorozko: All those changes should be in nineteenth commit, no?
The one that says:

> Refactor worker RX ports
> 
> Change unnamed embedded struct describing RX worker port to Endpoint
> Change container of RX worker ports to StaticVector

dataplane/controlplane.cpp
```diff
@@ -395,10 +374,8 @@ common::idp::get_ports_stats::response cControlPlane::get_ports_stats()
 		}
 
 		uint64_t physicalPort_egress_drops = 0;
-		for (const auto& [coreId, worker] : dataPlane->workers)
+		for (const cWorker* worker : dataPlane->workers_vector)
```
ol-imorozko: same here

dataplane/controlplane.cpp
```diff
@@ -579,10 +556,8 @@ common::idp::getAclCounters::response cControlPlane::getAclCounters()
 	common::idp::getAclCounters::response response;
 
 	response.resize(YANET_CONFIG_ACL_COUNTERS_SIZE);
-	for (const auto& [coreId, worker] : dataPlane->workers)
+	for (const cWorker* worker : dataPlane->workers_vector)
```
ol-imorozko, line 585: same here

dataplane/controlplane.cpp
```diff
@@ -646,9 +621,8 @@ common::idp::getCounters::response cControlPlane::getCounters(const common::idp:
 		}
 
 		uint64_t counter = 0;
-		for (const auto& [core_id, worker] : dataPlane->workers)
+		for (const cWorker* worker : dataPlane->workers_vector)
```
ol-imorozko: same here
and so on...

dataplane/dataplane.cpp
```diff
@@ -1343,8 +1318,47 @@ void cDataPlane::start()
...
+	for (auto& [core, worker] : workers)
+	{
+		if (coreFunctions_.find(core) != coreFunctions_.end())
+		{
+			YANET_LOG_ERROR("Worker: Multiple workloads assigned to core %d\n", core);
+		}
+		YANET_LOG_INFO("Worker assigned to core %d\n", core);
+		coreFunctions_.emplace(core, [worker]() {
+			worker->start();
+		});
```
ol-imorozko: Those are the changes for the separate commit. This is certainly not about "removing Removed runBarrier synchronization"

ol-imorozko (changes requested):
twenty-first commit:
Again, present simple in commit title: `Remove startup (runBarrier) synchronization`

Also, could you add a description explaining why we can remove that synchronization? 
I'm guessing it's the same reason as described in previous commit, but I'm not sure

dataplane/dataplane.h
```diff
@@ -104,6 +104,7 @@ class cDataPlane
 	{
 		return current_time;
 	}
+	std::string InterfaceNameFromPort(tPortId id) { return std::get<0>(ports[id]); };
```
ol-imorozko, line 122: `const`

ol-imorozko (changes requested):
third commit:

dataplane/kernel_interface_handler.h
```diff
@@ -0,0 +1,277 @@
...
+		{
+			rte_pktmbuf_free_bulk(m_burst + sent, remain);
+		}
+		m_burst_length = 0;
+	}
+	struct DirectionStats
+	{
+		uint64_t bytes;
+		uint64_t packets;
+		uint64_t dropped;
```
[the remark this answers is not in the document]
TheRandomCharacter: `sKniStats` are long lived structs tracking cumulative stats, initializing those on creation is needed and prefferabe.
`DirectionStats` are not cumulative, they are used for returning statistics on each transmission. Zero initializing them each time before overwriting with actual values is meaningless, although compiler should optimize this out, so why not.

dataplane/kernel_interface_handler.h
```diff
@@ -0,0 +1,277 @@
...
+			return total + rte_pktmbuf_pkt_len(mbuf);
+		});
+		stats.packets = rte_eth_tx_burst(m_port, m_queue, m_burst, m_burst_length);
+
+		stats.dropped = m_burst_length - stats.packets;
+		if (stats.dropped)
+		{
+			stats.bytes = std::accumulate(m_burst, m_burst + m_burst_length, stats.bytes, [](uint64_t total, rte_mbuf* mbuf) {
+				return total - rte_pktmbuf_pkt_len(mbuf);
+			});
```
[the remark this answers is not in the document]
TheRandomCharacter: We are deducting from `stats.bytes` lengths of packets that were not transmitted hence the starting value.
The first non transmitted packet is at `m_burst + stats.packets`, good catch.

dataplane/kernel_interface_handler.h
```diff
@@ -0,0 +1,277 @@
...
+private:
+	std::size_t size_ = 0;
+	PortArray<tPortId> phy_ports_;
+	PortArray<tQueueId> phy_queues_;
+	PortArray<sKniStats> stats_;
+	PortArray<KernelInterface> forward_;
+	PortArray<KernelInterface> in_dump_;
+	PortArray<KernelInterface> out_dump_;
+	PortArray<KernelInterface> drop_dump_;
+	const dataplane::base::PortMapper* port_mapper_;
```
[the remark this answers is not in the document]
TheRandomCharacter: Only first `size_` elements of `PortArray`s are used. Default member initializer is used to set `size_` to `0`. What is the benefit of initializing the remainder of unused elements?

dataplane/controlplane.cpp
```diff
@@ -351,11 +323,19 @@ common::idp::get_worker_gc_stats::response cControlPlane::get_worker_gc_stats()
 
 common::idp::get_dregress_counters::response cControlPlane::get_dregress_counters()
 {
-	auto guard = dregress.LockCounters();
+	common::dregress::counters_t counters_v4;
+	common::dregress::counters_t counters_v6;
+	for (auto it : dataPlane->slow_workers)
+	{
```
ol-imorozko, line 329: use reference

dataplane/dataplane.cpp
```diff
@@ -443,6 +450,7 @@ eResult cDataPlane::initPorts()
 		tPortId portId;
 		if (strncmp(name.data(), SOCK_DEV_PREFIX, strlen(SOCK_DEV_PREFIX)) == 0)
 		{
+			YANET_LOG_INFO("Opening sockdev with path %s\n", name.data() + strlen(SOCK_DEV_PREFIX));
 			portId = sock_dev_create(name.data() + strlen(SOCK_DEV_PREFIX), interfaceName.c_str(), 0);
 		}
 		else if (rte_eth_dev_get_port_by_name(name.data(), &portId))
```
ol-imorozko, line 456: oh no, you add integer to a `char*` here which shifts it. I understand why you do this, but we can do it so much better:

First of all, let's use `string::rfind` with an offset of 0 to check if the name starts with the prefix.
Then, instead of shifting the pointer, we can just create a new string.
Or even better, just create a non-owning `std::string_view`:
```cpp
        tPortId portId = 0;
        std::string_view name_view(name);

        if (name_view.starts_with(SOCK_DEV_PREFIX))
        {
            std::string_view sockdev_path = name_view.substr(strlen(SOCK_DEV_PREFIX));
            YANET_LOG_INFO("Opening sockdev with path %.*s\n", static_cast<int>(sockdev_path.size()), sockdev_path.data());
            portId = sock_dev_create(sockdev_path.data(), interfaceName.c_str(), 0);
        }
        else if (rte_eth_dev_get_port_by_name(name.c_str(), &portId) != 0)
        {
        //...
```

ol-imorozko (changes requested):
Last commit:
Some description to the commit will be nice, also add verb in the title, please:
`Implement multiple slow workers`

dataplane/dataplane.cpp
```diff
@@ -1242,51 +1335,56 @@ void cDataPlane::init_worker_base()
...
 		{
-			for (const auto& [socket_id, globalbase_atomic] : globalBaseAtomics)
+			// the only place thread-shared variable icmpPacketsToSW is changed
+			for (cWorker* worker : workers_vector)
 			{
-				(void)socket_id;
-				globalbase_atomic->currentTime = current_time;
+
+				__atomic_store_n(&worker->packetsToSWNPRemainder, config.SWNormalPriorityRateLimitPerWorker, __ATOMIC_RELAXED);
 			}
```
ol-imorozko, line 1378: Ugh. Let's use `std::atomic` with `std::memory_order_relaxed`.

Change declaration of packetsToSWNPRemainder to:
```cpp
// will decrease with each new packet sent to slow worker, replenishes each N mseconds
 std::atomic<int32_t> packetsToSWNPRemainder; 
```

And use here
```cpp
worker->packetsToSWNPRemainder.store(config.SWNormalPriorityRateLimitPerWorker, std::memory_order_relaxed);
```

dataplane/dataplane.cpp
```diff
@@ -1242,51 +1335,56 @@ void cDataPlane::init_worker_base()
...
+void cDataPlane::SWRateLimiterTimeTracker()
 {
-	uint32_t prev_time = 0;
-
 	for (;;)
 	{
-		current_time = time(nullptr);
+		using namespace std::chrono_literals;
+		// seem to be sufficiently fast function for slowWorker whose threshold is 200'000 packets per second
+		std::chrono::high_resolution_clock::time_point curTimePointForSWRateLimiter = std::chrono::high_resolution_clock::now();
```
ol-imorozko, line 1368: `auto curTimePointForSWRateLimiter = std::chrono::high_resolution_clock::now();`

dataplane/dataplane.cpp
ol-imorozko: Note that we also will need to change the following function:
```cpp
inline void cWorker::slowWorker_entry_normalPriority(rte_mbuf* mbuf,
                                                     const common::globalBase::eFlowType& flowType)
{
	/// @todo: worker::tStack

	if (basePermanently.SWNormalPriorityRateLimitPerWorker != 0 && __atomic_fetch_sub(&packetsToSWNPRemainder, 1, __ATOMIC_RELAXED) <= 0)
	{
		rte_pktmbuf_free(mbuf);
		counters[(uint32_t)common::globalBase::static_counter_type::slow_worker_normal_priority_rate_limit_exceeded]++;

		return;
	}

	dataplane::metadata* metadata = YADECAP_METADATA(mbuf);
	metadata->flow.type = flowType;

	if (rte_ring_sp_enqueue(ring_normalPriority, (void*)mbuf))
	{
		stats.ring_normalPriority_drops++;
		rte_pktmbuf_free(mbuf);
	}
	else
	{
		stats.ring_normalPriority_packets++;
	}
}
```
to
```cpp
inline void slowWorker_entry_normalPriority(rte_mbuf* mbuf, const common::globalBase::eFlowType& flowType)
{
        /// @todo: worker::tStack

        if (basePermanently.SWNormalPriorityRateLimitPerWorker != 0 && packetsToSWNPRemainder.fetch_sub(1, std::memory_order_relaxed) <= 0)
        {
            rte_pktmbuf_free(mbuf);
            counters[static_cast<uint32_t>(common::globalBase::static_counter_type::slow_worker_normal_priority_rate_limit_exceeded)]++;
            return;
        }

        auto* metadata = YADECAP_METADATA(mbuf);
        metadata->flow.type = flowType;

        if (rte_ring_sp_enqueue(ring_normalPriority, static_cast<void*>(mbuf)) != 0)
        {
            stats.ring_normalPriority_drops++;
            rte_pktmbuf_free(mbuf);
        }
        else
        {
            stats.ring_normalPriority_packets++;
        }
    }
```

dataplane/dataplane.cpp
```diff
@@ -1335,22 +1432,24 @@ void cDataPlane::start()
...
-	if (coreFunctions_.find(config.controlPlaneCoreId) != coreFunctions_.end())
+	for (auto& [core, slow] : slow_workers)
 	{
-		YANET_LOG_ERROR("CP: Multiple workloads assigned to core %d\n", config.controlPlaneCoreId);
+		if (coreFunctions_.find(core) != coreFunctions_.end())
+		{
+			YANET_LOG_ERROR("Multiple workloads assigned to core %d\n", core);
+		}
+		coreFunctions_.emplace(core, [slow]() {
+			auto work_runner = dpdk::WorkRunner{slow, dpdk::Yielder{}};
```
ol-imorozko, line 1449: Captured structured bindings are a C++20 extension, we cannot use them yet

ol-imorozko (changes requested):
Last commit:

dataplane/slow_worker.h
```diff
@@ -1,22 +1,147 @@
...
+	           const std::vector<tPortId>& ports_to_service,
+	           const std::vector<cWorker*>& workers_to_service,
+	           std::vector<dpdk::RingConn<rte_mbuf*>>&& from_gcs,
+	           KernelInterfaceWorker&& kni,
+	           rte_mempool* mempool,
+	           bool use_kni,
+	           uint32_t sw_icmp_out_rate_limit);
+	SlowWorker(const SlowWorker&) = delete;
+	SlowWorker(SlowWorker&& other);
+
```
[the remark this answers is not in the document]
ol-imorozko: Also we should add move assignment operator

dataplane/slow_worker.h
```diff
@@ -1,22 +1,147 @@
...
+	void WaitInit();
+	void Start()
+	{
+		WaitInit();
+		while (!m_stop)
+		{
+			Iteration();
+		}
+	}
```
ol-imorozko, line 95: I couldn't find any code that changes `m_stop`. So this seems like an infinite loop for me

dataplane/slow_worker.h
```diff
@@ -1,22 +1,147 @@
...
+				m_slow_worker_mbufs.pop();
+			}
+
+			m_slow_worker->slowWorkerHandlePackets();
+		}
+
+		m_slow_worker->slowWorkerAfterHandlePackets();
+#ifdef CONFIG_YADECAP_AUTOTEST
+		std::this_thread::sleep_for(std::chrono::microseconds{1});
+#endif // CONFIG_YADECAP_AUTOTEST
```
ol-imorozko, line 143: Do we need a mutex here? I'm not sure whether this code could be executed from different thread, but if it does, then `m_slow_worker_mbufs` seems like a shared resource worth protecting

dataplane/slow_worker.cpp
```diff
@@ -1,25 +1,968 @@
...
+        m_mempool{other.m_mempool},
+        m_fragmentation{std::move(other.m_fragmentation)},
+        m_dregress{std::move(other.m_dregress)},
+        m_kni_worker{std::move(other.m_kni_worker)}
+{
+	m_fragmentation.Callback() = [this](rte_mbuf* pkt, const common::globalBase::tFlow& flow) {
+		SendToSlowWorker(pkt, flow);
+	};
+}
+
```
ol-imorozko, line 48: What is the difference between this and the default move constructor? It appears we are simply moving all fields, which is the behavior of the default move constructor. Is it necessary to explicitly set the `Callback()` to `SendToSlowWorker(pkt, flow)` after the move? 
I mean ff the `Callback()` could potentially point to a different function, changing it after the move seems strange and could imply we are not just moving an object but also modifying its state

dataplane/slow_worker.cpp
```diff
@@ -1,25 +1,968 @@
...
+	unsigned rxSize = rte_ring_sc_dequeue_burst(ring,
+	                                            (void**)mbufs,
+	                                            CONFIG_YADECAP_MBUFS_BURST_SIZE,
+	                                            nullptr);
+
+#ifdef CONFIG_YADECAP_AUTOTEST
+	if (rxSize)
+	{
+		std::this_thread::sleep_for(std::chrono::microseconds{400});
+	}
```
ol-imorozko, line 127: magic value

dataplane/work_runner.h
```diff
@@ -0,0 +1,79 @@
...
+	{
+	}
+};
+
+template<typename... Workloads>
+WorkRunner<Workloads...> MakeWorkRunner(Workloads&&... loads)
+{
+	return WorkRunner<Workloads...>{std::forward(loads)...};
+}
+
```
ol-imorozko, line 60: I don't think this will compile. Using `std::forward` without specifying the template type seems incorrect. How is it supposed to know whether to use the copy or move constructors? `MakeWorkRunner` is an unused function, which might be why we haven't encountered compilation errors yet.

dataplane/worker.h
```diff
@@ -217,15 +216,21 @@ class cWorker
...
 	YANET_NEVER_INLINE void slowWorkerHandleFragment(rte_mbuf* mbuf);
 	YANET_NEVER_INLINE void slowWorkerFarmHandleFragment(rte_mbuf* mbuf);
+
 	YANET_NEVER_INLINE void slowWorkerAfterHandlePackets();
+
 	YANET_NEVER_INLINE void slowWorkerFlow(rte_mbuf* mbuf, const common::globalBase::tFlow& flow);
 	YANET_NEVER_INLINE void slowWorkerTranslation(rte_mbuf* mbuf, const dataplane::globalBase::tNat64stateless& nat64stateless, const dataplane::globalBase::nat64stateless_translation_t& translation, bool direction); /** true: ingress, false: egress */
+	const dataplane::base::generation& CurrentBase() { return bases[localBaseId & 1]; }
+	void IncrementCounter(common::globalBase::static_counter_type type) { counters[(uint32_t)type]++; }
+	uint32_t CurrentTime() { return basePermanently.globalBaseAtomic->currentTime; }
```
ol-imorozko, line 233: `const`

dataplane/worker.h
```diff
@@ -217,15 +216,21 @@ class cWorker
...
 	YANET_NEVER_INLINE void slowWorkerHandlePackets();
+
 	YANET_NEVER_INLINE void slowWorkerHandleFragment(rte_mbuf* mbuf);
 	YANET_NEVER_INLINE void slowWorkerFarmHandleFragment(rte_mbuf* mbuf);
+
 	YANET_NEVER_INLINE void slowWorkerAfterHandlePackets();
+
 	YANET_NEVER_INLINE void slowWorkerFlow(rte_mbuf* mbuf, const common::globalBase::tFlow& flow);
 	YANET_NEVER_INLINE void slowWorkerTranslation(rte_mbuf* mbuf, const dataplane::globalBase::tNat64stateless& nat64stateless, const dataplane::globalBase::nat64stateless_translation_t& translation, bool direction); /** true: ingress, false: egress */
+	const dataplane::base::generation& CurrentBase() { return bases[localBaseId & 1]; }
```
ol-imorozko, line 230: `const`

dataplane/work_runner.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
+#include <chrono>
+#include <iostream>
```
ol-imorozko, line 3: `iostream` is unused

dataplane/work_runner.h
```diff
@@ -0,0 +1,79 @@
...
+}
+
+} // namespace internal
+
+template<typename... Workloads>
+class WorkRunner
+{
+	std::tuple<Workloads...> m_workers;
+	bool m_running = true;
+	template<int I = 0>
```
ol-imorozko, line 31: who changes `m_running`?

ol-imorozko (dismissed):
last commit:

dataplane/dataplane.cpp
```diff
@@ -534,6 +569,112 @@ eResult cDataPlane::initPorts()
...
+	}
+
+	if (config.use_kernel_interface)
+	{
+		for (auto& it : kni_interface_handles)
+		{
+			if (!it.second.forward.SetUp())
+			{
+				YANET_LOG_ERROR("Failed to set kni interface belonging to %s up", std::get<0>(ports.at(it.first)).c_str());
+				std::terminate();
```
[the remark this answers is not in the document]
TheRandomCharacter: Consistency is a valid point.
`std::abort` is recommended over `std::terminate` in cases where additional functionality of `std::terminate` is not used, so `std::abort` is preferred. Maybe we should even consider switching to using `std::exit` for such cases.

dataplane/dataplane.cpp
```diff
@@ -534,6 +569,112 @@ eResult cDataPlane::initPorts()
...
+		rte_eth_promiscuous_enable(portId);
+	}
+
+	if (config.use_kernel_interface)
+	{
+		for (auto& it : kni_interface_handles)
+		{
+			if (!it.second.forward.SetUp())
+			{
+				YANET_LOG_ERROR("Failed to set kni interface belonging to %s up", std::get<0>(ports.at(it.first)).c_str());
```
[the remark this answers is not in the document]
TheRandomCharacter: don't get the purpose of `std::move` in this context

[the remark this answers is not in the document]
ol-imorozko: Oh, I don't know how it get there, sorry. Yeah, this is solely about using structured bindings instead of `it.second` and `it.first`

dataplane/kernel_interface_handle.cpp
```diff
@@ -0,0 +1,157 @@
...
+{
+	*this = std::move(other);
+}
```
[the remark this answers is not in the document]
TheRandomCharacter: I would rather sacrifice efficiency for code deduplication here, since speed isn't critical for `KernelInterfaceHandle` that is long lived.

dataplane/fragmentation.cpp
TheRandomCharacter: Same here, duplicationg logic is IMHO more error prone

common/ctree.h
```diff
@@ -24,6 +26,19 @@ class ctree
...
+		if (this != &other)
+		{
+			*this = std::move(other);
+		}
+	}
+	ctree& operator=(ctree&& other)
+	{
+		std::swap(root_node, other.root_node);
+		return *this;
+	}
```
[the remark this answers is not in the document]
TheRandomCharacter: `root_node` is a raw owning pointer, copying it as suggested will result in it being set up for double `delete`

common/ctree.h
TheRandomCharacter: even better we can define this operator for `ctree<Size, Counters...>::values_t`

common/ctree.h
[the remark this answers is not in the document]
ol-imorozko: Oh, yea, you're right, I meant to add `std::exchange` there but forgot.
I think smth like that will do:
```cpp
ctree(ctree&& other) noexcept :
        root_node(std::exchange(other.root_node, nullptr))
    {}

    ctree& operator=(ctree&& other) noexcept
    {
        if (this != &other)
        {
            root_node = std::exchange(other.root_node, nullptr);
        }
        return *this;
    }
```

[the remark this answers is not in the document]
ol-imorozko: or `std::swap` instead of `std::exchange`, yeah, doesn't matter much there

common/ctree.h
TheRandomCharacter: nevermind, node types are dependent and that interferes with providing `operator+=` for `ctree::values_t`

dataplane/dregress.h
TheRandomCharacter: and i would have thought the main problem is that both return v4 counters...

dataplane/dregress.cpp
TheRandomCharacter: I agree, but this doesn't relate to the theme of commit. This should be done in separate commit.

dataplane/dataplane.cpp
TheRandomCharacter: This can of worms is better handled in separate refactoring PR

dataplane/work_runner.h
TheRandomCharacter: Currently no one. Removed

dataplane/globalbase.cpp
```diff
@@ -2391,39 +2402,45 @@ eResult generation::dregress_local_prefix_update(const common::idp::updateGlobal
...
 		{
-			dataPlane->controlPlane->dregress.local_prefixes_v6.emplace(prefix.get_ipv6());
+			if (prefix.is_ipv4())
+			{
+				dregress.local_prefixes_v4.emplace(prefix.get_ipv4());
+			}
+			else
+			{
+				dregress.local_prefixes_v6.emplace(prefix.get_ipv6());
+			}
```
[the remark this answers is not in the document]
TheRandomCharacter: `insert`, since it'a a set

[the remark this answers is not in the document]
TheRandomCharacter: prefix is part of request that we are getting by `const &`, moving it is not a good idea

dataplane/slow_worker.h
TheRandomCharacter: Start() is also unused, removed

dataplane/slow_worker.h
TheRandomCharacter: No mutex needed, `m_slow_worker_mbufs` is used by SlowWorker for stashing packets for later handling. It's filled by `SendToSlowWorker`.

dataplane/slow_worker.cpp
TheRandomCharacter: `Callback()` lambda is capturing `this` pointer to SlowWorker. Since we are moving SlowWorker the old pointer passed in this lambda to `Fragmentation` becomes invalid after move.

dataplane/controlplane.h
```diff
@@ -163,24 +135,48 @@ class cControlPlane ///< @todo: move to cDataPlane
...
+			Data* operator->() const noexcept { return &data_; }
+			Data& operator*() const { return data_; }
+		};
+
+	public:
+		SequentialAccessor Accessor()
+		{
+			return SequentialAccessor{mutex_, data_};
+		}
+	};
```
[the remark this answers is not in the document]
TheRandomCharacter: Passing an op as callback does seem cleaner but arguably harder to understand. For example:
```c++
	auto vip_to_balancers = slow_worker_->dataPlane->controlPlane->vip_to_balancers.Accessor();
	auto vip_vport_proto = slow_worker_->dataPlane->controlPlane->vip_vport_proto.Accessor();
       ...
```
will transform into nested lambdas.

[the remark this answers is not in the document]
TheRandomCharacter: changed to passing critical sections as arguments

dataplane/controlplane.cpp
```diff
@@ -272,29 +242,31 @@ common::idp::getWorkerStats::response cControlPlane::getWorkerStats(const common
 
 		for (const cWorker* worker : dataPlane->workers_vector)
 		{
```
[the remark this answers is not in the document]
TheRandomCharacter: I would prefer we don't use auto when type is known and name is short. Changed other `auto`s to `cWorker*`

GeorgyKirichenko (approved):

ol-imorozko:
@GeorgyKirichenko @TheRandomCharacter 

> let us fix the rest in separate piece of work

It's important not to forget to do this. Most of the remaining review notes are minor, but there are also some significant ones: https://github.com/yanet-platform/yanet/pull/217#discussion_r1684206548

PR #230, opened 2024-07-26
Fixes route.v4.lpm and route.v6.lpm reallocation on update. Adds interfacesRequired config flag

TheRandomCharacter:


ol-imorozko (changes requested):
For the first commit:

Title is too long, I suggest `Abort startup if no interfaces and 'interfacesRequired' flag is set`

dataplane/memory_manager.h
```diff
@@ -42,6 +42,44 @@ class memory_manager
...
+	{
+		return std::unique_ptr<type, Deleter>{
+		        reinterpret_cast<type*>(alloc(name,
+		                                      socket_id,
+		                                      size,
+		                                      [](void* pointer) {
+			                                      reinterpret_cast<type*>(pointer)->~type();
+		                                      })),
+		        Deleter{this}};
+	}
```
ol-imorozko, line 81: I'm not sure about that.. I think we can just use custom deleter directly with `std::unique_ptr` and have the same functionality by using simpler constructs.

```cpp
template<typename T, typename... Args>
std::unique_ptr<T, std::function<void(T*)>> create_unique(const char* name,
                                                          tSocketId socket_id,
                                                          size_t size,
                                                          Args&&... args)
{
	T* raw_pointer = alloc(name, socket_id, size, [](void* ptr) { reinterpret_cast<T*>(ptr)->~T(); });
	return std::unique_ptr{raw_pointer, [this](T* ptr) { this->destroy(ptr); }};

}
```

dataplane/updater.h
```diff
@@ -160,161 +142,69 @@ class updater_lpm4_24bit_8bit
...
-	eResult init()
-	{
-		return create(2 * object_type::extended_chunks_size_min);
-	}
+private:
+	std::string name;
+	dataplane::memory_manager* memory_manager;
+	tSocketId socket_id;
+	stats_t stats;
+	dataplane::memory_manager::unique_ptr<ObjectType> pointer_;
```
ol-imorozko, line 161: `    std::unique_ptr<ObjectType, std::function<void(ObjectType*)>> pointer_;`

dataplane/updater.h
```diff
@@ -160,161 +142,69 @@ class updater_lpm4_24bit_8bit
...
-			memory_manager->destroy(pointer);
-		}
-
+		next->copy(next_stats, stats, *(pointer_.get()));
 		stats = next_stats;
-		pointer = next_pointer;
-
+		std::swap(pointer_, next);
 		return eResult::success;
 	}
```
ol-imorozko, line 182: And here we can do it like:
```cpp
eResult create(size_t size) {
        auto new_pointer = memory_manager->create_unique<ObjectType>(name.c_str(), socket_id, ObjectType::calculate_sizeof(size));
        if (!new_pointer) {
            return eResult::errorAllocatingMemory;
        }

        stats_t new_stats;
        new_stats.extended_chunks_count = 0;
        new_stats.extended_chunks_size = size;
        new_stats.max_used_chunk_id = 0;
        new_stats.free_chunk_cache.flags = 0;

        if (pointer_) {
            new_pointer->copy(new_stats, stats, *pointer_);
        }

        stats = new_stats;
        pointer_ = std::move(new_pointer);
        return eResult::success;
    }
```

```cpp
    eResult Resize(size_t new_size) {
        return create(new_size);
    }
```

dataplane/updater.h
```diff
@@ -38,98 +38,78 @@ namespace dataplane
...
-			next_pointer->copy(next_stats, stats, *pointer);
-			memory_manager->destroy(pointer);
+			return eResult::errorAllocatingMemory;
 		}
-
-		stats = next_stats;
-		pointer = next_pointer;
-
-		return eResult::success;
 	}
```
ol-imorozko, line 74: Let's do it like in `updater_array` to be consistent, i.e reintroduce the `create` method to allocate memory and set up the `pointer_`, and here just do
```cpp
    eResult init() {
        return create(2 * ObjectType::extended_chunks_size_min);
    }
```

ol-imorozko (changes requested):
first commit:

dataplane/memory_manager.h
TheRandomCharacter: Can't seem to see how using `std::function` results in simpler constructs here. Except for moving raw_pointer allocation out of return statement those are virtually identical.

On the other hand using `std::function` means that a deleter has to store not only pointer to `memory_manager` but also a pointer for vtabe. This would double deleter's memory footprint.

dataplane/updater.h
TheRandomCharacter: I really dislike that `create(size)` has a hidden logic of copying data without any mention or hint at call site.
Addition of separate `Resize(size)` is made specifically because of this.

dataplane/memory_manager.h
ol-imorozko: Yeah, you right. The overhead of `std::function` is indeed larger than that of a custom deleter class.
let's just move raw pointer allocation out then and do other minor corrections:
- use typename `T, Args` instead of `type, args_t`
- remove unnecessary const
- `Args&&` instead of `const args_t&`



Ehh.. we could speculate a little about using `auto` as a return type auto so the compiler will use the lambda directly as a callable object, and a std::function will not be created.
But since we need to store that unique_ptr as a field in another class, we will need the type. So we can extract lambda to a function with `auto` as a return type and then use `std::result_of_t`. Smth like this:
```cpp
    auto CreateDeleter() {
        return [this](auto* ptr) {
            this->destroy(ptr);
        };
    }
   using Deleter = std::result_of_t<decltype(&memory_manager::CreateDeleter)>;

    template<typename T, typename... Args>
    auto create_unique(const char* name, tSocketId socket_id, size_t size, Args&&... args) {
        T* raw_pointer = reinterpret_cast<T*>(alloc(name, socket_id, size, [](void* ptr) {
            reinterpret_cast<T*>(ptr)->~T();
        }));
        return std::unique_ptr(raw_pointer, CreateDeleter());
    }
    //...
    using Deleter = std::result_of_t<decltype(&memory_manager::CreateDeleter)>;
```

But I don't think that this is better than just a custom class-deleter.

dataplane/updater.h
ol-imorozko: Ok, we don't want to reintroduce the `create` since it has some hidden logic of copying data, but we sill have some code duplication with `init` and `Resize` methods. Maybe some other private method (not `create`) will do the job?

dataplane/memory_manager.h
ol-imorozko: oh, we can do that even without a function, just call `::deleter_type` from `unique_ptr`. Didn't know that

dataplane/updater.h
TheRandomCharacter: How about
```cpp
	memory_manager::unique_ptr<ObjectType> Allocate(const std::size_t size)
	{
		return memory_manager->create_unique<ObjectType>(name.c_str(),
		                                                 socket - id,
		                                                 ObjectType::calculate_sizeof(size));
	}
```
Everything else is specific to `init` and `Resize`

TheRandomCharacter: `updater_array` is different not only in it's "arrayness", but also doesn't track stats. IMHO naming similarity is unfortunate coincidence.

dataplane/memory_manager.h
```diff
@@ -42,6 +42,45 @@ class memory_manager
...
+	template<typename T>
+	using unique_ptr = std::unique_ptr<T, Deleter>;
+
+	template<typename T,
+	         typename... Args>
+	unique_ptr<T> create_unique(const char* name,
+	                            const tSocketId socket_id,
+	                            const uint64_t size,
+	                            const Args&... args)
+	{
```
ol-imorozko, line 72: `Args&&... args`

dataplane/memory_manager.h
```diff
@@ -42,6 +42,45 @@ class memory_manager
...
+		void* pointer = alloc(name,
+		                      socket_id,
+		                      size,
+		                      [](void* pointer) {
+			                      reinterpret_cast<T*>(pointer)->~T();
+		                      });
+		return std::unique_ptr<T, Deleter>{
+		        new (pointer) T(args...),
```
ol-imorozko, line 80: `new (pointer) T(std::forward<Args>(args)...),`

ol-imorozko (approved):

ol-imorozko (approved):

dataplane/updater.h
```diff
@@ -38,98 +38,80 @@ namespace dataplane
...
-		if (stats.extended_chunks_size - stats.extended_chunks_count < object_type::extended_chunks_size_min)
+		if (NeedToGrow())
 		{
-			result = create(stats.extended_chunks_size * 2);
+			eResult result = Resize(GrowSize());
 			if (result != eResult::success)
 			{
 				return result;
 			}
 		}
```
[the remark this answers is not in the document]
TheRandomCharacter: This is legacy logic. Seemed strange to me too. let's remove it.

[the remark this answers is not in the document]
TheRandomCharacter: Actually no. it's an lpm. Removing an ip may split a node representing iprange into two, increasing lpm size.

GeorgyKirichenko (approved):

PR #234, opened 2024-08-06
Fixes UB when buffersize is zero

TheRandomCharacter:
Passing invalid pointer to memcpy is UB even if `bufferSize` is zero.

### DIFF
--- a/common/stream.h
+++ b/common/stream.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <inttypes.h>
-#include <string.h>
 
 #include <map>
 #include <memory>
@@ -28,12 +27,18 @@ public:
 	template<typename TType>
 	inline void pop(TType& value);
 
+private:
 	inline void pop(char* buffer, uint64_t bufferSize);
+
+public:
 	inline void pop(std::string& value);
 	inline void pop(std::vector<uint8_t>& value);
 
 	template<typename TFirst, typename TSecond>
 	inline void pop(std::pair<TFirst, TSecond>& pair);
+
+	template<typename T, std::size_t Size>
+	inline void pop(T (&value)[Size]);
 
 	template<typename TType, std::size_t TSize>
 	inline void pop(std::array<TType, TSize>& array);
@@ -93,7 +98,10 @@ public:
 	template<typename TType>
 	inline void push(const TType& value);
 
+private:
 	inline void push(const char* buffer, uint64_t bufferSize);
+
+public:
 	inline void push(const std::string& value);
 	inline void push(const std::vector<uint8_t>& value);
 
@@ -103,6 +111,15 @@ public:
 	{
 		push(pair.first);
 		push(pair.second);
+	}
+
+	template<typename T, std::size_t Size>
+	inline void push(const T (&array)[Size])
+	{
+		for (const auto& e : array)
+		{
+			push(e);
+		}
 	}
 
 	template<typename TType, std::size_t TSize>
@@ -273,19 +290,17 @@ inline stream_in_t::stream_in_t(const std::vector<uint8_t>& buffer) :
 template<typename TType>
 inline void stream_in_t::pop(TType& value)
 {
-	if constexpr (std::is_trivial_v<TType> &&
-	              !(std::is_pointer_v<TType> ||
-	                std::is_reference_v<TType>))
+	if constexpr (std::is_trivially_copyable_v<TType>)
 	{
 		if (inBuffer.size() - inPosition < sizeof(TType))
 		{
-			memset(&value, 0, sizeof(TType));
+			value = TType{};
 			inPosition = this->inBuffer.size();
 			failed = true;
 			return;
 		}
 
-		memcpy(&value, &inBuffer[inPosition], sizeof(TType));
+		value = reinterpret_cast<const TType&>(inBuffer[inPosition]);
 
 		inPosition += sizeof(TType);
 	}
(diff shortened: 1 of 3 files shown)


ol-imorozko (approved):

common/stream.h
```diff
@@ -525,13 +557,13 @@ inline stream_out_t::stream_out_t()
 template<typename TType>
 inline void stream_out_t::push(const TType& value)
 {
-	if constexpr (std::is_trivial_v<TType> &&
+	if constexpr (std::is_trivially_copyable_v<TType> &&
 	              !(std::is_pointer_v<TType> ||
 	                std::is_reference_v<TType>))
```
ol-imorozko, line 562: Could you also remove this part: ` !(std::is_pointer_v<TType> || std::is_reference_v<TType>)` due to it being redundant?

Template type deduction cannot result in a `TType` being a reference or a pointer type, since `value` is not of a a universal reference type

common/stream.h
```diff
@@ -273,19 +291,19 @@ inline stream_in_t::stream_in_t(const std::vector<uint8_t>& buffer) :
...
 		{
-			memset(&value, 0, sizeof(TType));
+			value = TType{};
 			inPosition = this->inBuffer.size();
 			failed = true;
 			return;
 		}
 
-		memcpy(&value, &inBuffer[inPosition], sizeof(TType));
+		value = reinterpret_cast<const TType&>(inBuffer[inPosition]);
```
ol-imorozko, line 306: Why is `memcpy` deemed worse than using `reinterpret_cast` to a reference for copying trivially copyable objects?

In such context both `memcpy` and `reinterpret_cast` should be of similar performance, compiler will translate them both to an operation like `mov`.

IMHO `std::memcpy` is more idiomatic for copying bytes from a buffer to a target variable in this scenario.

Also I would use `std::memcpy(&value, inBuffer.data() + inPosition, sizeof(TType));` instead of `&inBuffer[inPosition]` for a second argument

common/stream.h
```diff
@@ -525,13 +557,13 @@ inline stream_out_t::stream_out_t()
...
 {
-	if constexpr (std::is_trivial_v<TType> &&
+	if constexpr (std::is_trivially_copyable_v<TType> &&
 	              !(std::is_pointer_v<TType> ||
 	                std::is_reference_v<TType>))
 	{
 		integer_t size = outBuffer.size();
 		outBuffer.resize(size + sizeof(TType));
-		memcpy(&outBuffer[size], &value, sizeof(TType));
+		reinterpret_cast<TType&>(outBuffer[size]) = value;
```
ol-imorozko, line 571: Same thing here, IMHO
```cpp
if constexpr (std::is_trivially_copyable_v<TType>)
{
    auto data = reinterpret_cast<const uint8_t*>(&value);
    outBuffer.insert(outBuffer.end(), data, data + sizeof(TType));
}
```
is much more cleaner

common/stream.h
```diff
@@ -304,24 +322,29 @@ inline void stream_in_t::pop(char* buffer, uint64_t bufferSize)
...
-	pop(&value[0], size);
-	value[size] = 0;
+	if (this->inBuffer.size() - inPosition < size)
+	{
+		inPosition = this->inBuffer.size();
+		failed = true;
+		return;
+	}
+
+	value = std::string(reinterpret_cast<const char*>(&this->inBuffer[inPosition]), size);
```
ol-imorozko, line 351: `value.assign(reinterpret_cast<const char*>(inBuffer.data() + inPosition), size);`

ol-imorozko (changes requested):

common/stream.h
TheRandomCharacter: Our intent is not to copy some memory from one place to another, rather to treat position in buffer as `TType`.
Using `reinterpret_cast` reduces "byte magic" to a minimum, and states the change of type explicitly.

common/stream.h
TheRandomCharacter:
```c++
if constexpr (std::is_trivially_copyable_v<TType>)
{
	auto& data = reinterpret_cast<const uint8_t(&)[sizeof(TType)]>(value);
	outBuffer.insert(outBuffer.end(), std::begin(data), std::end(data));
}
```
IMHO is even cleaner

common/stream.h
```diff
@@ -304,21 +319,34 @@ inline void stream_in_t::pop(char* buffer, uint64_t bufferSize)
 		return;
 	}
 
-	memcpy(buffer, &this->inBuffer[inPosition], bufferSize);
+	if (bufferSize == 0)
+	{
+		return;
+	}
 
+	std::copy(&this->inBuffer[inPosition], &this->inBuffer[inPosition] + bufferSize, buffer);
```
ol-imorozko, line 327: why `this->inBuffer`? just `inBuffer` is enough.
Personally, I would use iterators with `std::copy`:
```cpp
    auto bufferStart = inBuffer.begin() + inPosition;

    std::copy(bufferStart, bufferStart + bufferSize, buffer);
```

common/stream.h
```diff
@@ -522,13 +559,10 @@ inline stream_out_t::stream_out_t()
...
-	if constexpr (std::is_trivial_v<TType> &&
-	              !(std::is_pointer_v<TType> ||
-	                std::is_reference_v<TType>))
+	if constexpr (std::is_trivially_copyable_v<TType>)
 	{
-		integer_t size = outBuffer.size();
-		outBuffer.resize(size + sizeof(TType));
-		memcpy(&outBuffer[size], &value, sizeof(TType));
+		auto& data = reinterpret_cast<const uint8_t(&)[sizeof(TType)]>(value);
+		outBuffer.insert(outBuffer.end(), std::begin(data), std::end(data));
```
ol-imorozko, line 565: _Feel free to ignore_

I become a little scared when I see a reference to a C-style array in an expression. I would recommend:
```cpp
    using ByteArray = const uint8_t(&)[sizeof(TType)];

    if constexpr (std::is_trivially_copyable_v<TType>)
    {
        auto& data = reinterpret_cast<ByteArray>(value);
        //...
 ```

ol-imorozko (dismissed):

GeorgyKirichenko (approved):

PR #242, opened 2024-09-08
VRF implementation for dataplane

stal76:
Previously, it was possible to add routing rules for different VRFs, but it was not possible to set VRFs for packets and only 'default' was used for them.
In the current change, the ability to set VRF for the entire logical port has been added. Later it will be possible to set VRF by acl rules.
To solve the LPM problem, an algorithm like to 'a path-compressed trie' is used.


common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	void Init(void* pointer, size_t size, Index number_first_reserved)
+	{
+		// size >= groups_number_ * (sizeof(Group) + 64 * BlockSize)
+		groups_number_ = size / (sizeof(Group) + 64 * BlockSize);
+		pointer_ = pointer;
+		blocks_ = pointer;
+		groups_ = reinterpret_cast<Group*>(reinterpret_cast<uint8_t*>(pointer) + groups_number_ * BlockSize * 64);
+
```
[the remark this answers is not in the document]
ol-imorozko: Oh, actually, you have a function for that: 
```cpp
groups_ = sdp::ShiftBuffer<Group*>(pointer, blocks_total_ * BlockSize);
```

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+			first_available_group_ = group_index;
+		}
+		blocks_used_--;
+		group.Free(block_index % 64);
+	}
+
+	void* operator[](Index block_index) const
+	{
+		return reinterpret_cast<uint8_t*>(blocks_) + block_index * BlockSize;
+	}
```
ol-imorozko, line 90: `return sdp::ShiftBuffer<void*>(blocks_, block_index * BlockSize);`

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+				Allocate();
+			}
+			blocks_used_ = number_first_reserved;
+		}
+	}
+
+	void* GetPointer() const
+	{
+		return pointer_;
+	}
```
ol-imorozko, line 55: mark `[[nodiscard]]`

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+		group.Free(block_index % 64);
+	}
+
+	void* operator[](Index block_index) const
+	{
+		return reinterpret_cast<uint8_t*>(blocks_) + block_index * BlockSize;
+	}
+
+	void Swap(BlocksAllocator<BlockSize>& other)
+	{
```
ol-imorozko, line 93: annotate with `noexcept`

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+
+	bool Empty() const
+	{
+		return pointer_ == nullptr;
+	}
+
+	std::pair<size_t, size_t> Stat() const
+	{
+		return {blocks_total_, blocks_used_};
+	}
```
ol-imorozko, line 111: mark with `[[nodiscard]]`

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+			return result;
+		}
+
+		void Free(uint8_t bit)
+		{
+			common::bits::enable_bit_64(available, bit);
+		}
+
+		uint64_t available;
+		Index next;
```
ol-imorozko, line 145: Use `std::bitset` for bit operations:
```cpp
        std::bitset<BitsInGroup> available;
        Index next = null_block;
```

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+		return {blocks_total_, blocks_used_};
+	}
+
+private:
+	struct Group
+	{
+		void Init(Index next)
+		{
+			available = static_cast<uint64_t>(-1);
+			this->next = next;
```
ol-imorozko, line 119: `available.set()`

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		bool Busy()
+		{
+			return available == 0;
+		}
```
ol-imorozko, line 125: `return available.none();`

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+
+		void Occupy()
+		{
+			available = 0;
```
ol-imorozko, line 129: `available.reset();`

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+		}
+
```
ol-imorozko, line 131: If my above proposal is correct, add
```cpp
void ReserveFirstN(uint8_t n)
{
	for (uint8_t i = 0; i < n; ++i)
	{
		available.reset(i);
	}
}
```

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+		uint8_t Allocate()
+		{
+			uint8_t result = common::bits::get_last_enabled_bit_64(available);
+			common::bits::disable_bit_64(available, result);
+			return result;
```
ol-imorozko, line 136:
```cpp
uint8_t first_available = available._Find_first();
available.reset(first_available);
return first_available;
```

common/blocks_allocator.h
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		void Free(uint8_t bit)
+		{
+			common::bits::enable_bit_64(available, bit);
```
ol-imorozko, line 141: `available.set(bit);`

common/unittest/bits.cpp
```diff
@@ -0,0 +1,409 @@
...
+	{
+		uint64_t mask = common::bits::build_mask_64(length);
+		ASSERT_EQ(common::bits::get_last_enabled_bit_64(mask), 64 - length);
+	}
+
+	Test32(common::bits::get_last_enabled_bit_32, TestGetLastEnabledBit<uint32_t>);
+	Test64(common::bits::get_last_enabled_bit_64, TestGetLastEnabledBit<uint64_t>);
+}
+
+} // namespace
```
ol-imorozko, line 409: Not needed

common/unittest/blocks_allocator.cpp
```diff
@@ -0,0 +1,70 @@
+#include <gtest/gtest.h>
+#include <memory>
+
+#include "../blocks_allocator.h"
+
+namespace
+{
+
```
ol-imorozko, line 8: These are good. Though I would add a test that covers a case where `number_first_reserved` gets equal to `blocks_total_`, smth like this should work:
```cpp
TEST(BlocksAllocator, SequentialAllocationFull)
{
	size_t blocks_number = 1024;
	uint32_t reserved = blocks_number;

	size_t size = common::allocator::BlocksAllocator<64>::GetBufferSize(blocks_number);
	std::unique_ptr<uint8_t[]> buffer = std::make_unique<uint8_t[]>(size);
	common::allocator::BlocksAllocator<64> allocator;
	allocator.Init(buffer.get(), size, reserved);

	// no more blocks shoudl be available
	ASSERT_EQ(allocator.Allocate(), common::allocator::BlocksAllocator<64>::null_block);
	ASSERT_EQ(allocator.Stat(), std::make_pair(blocks_number, blocks_number));

	for (size_t i = 0; i < reserved; i++)
	{
		ASSERT_EQ(allocator[i], buffer.get() + 64 * i);
	}
}
```

ol-imorozko (changes requested):

dataplane/vrf.h
```diff
@@ -0,0 +1,91 @@
...
+		VrfLpm4Linear::BlocksAllocator::Index next_block;
+		uint32_t ipAddress;
+		uint32_t valueId;
+		uint8_t mask;
+		uint8_t used;
+	};
+	static_assert(sizeof(OneBlock) <= size_of_chunk, "invalid size of OneBlock");
+
+	BlocksAllocator& allocator_;
+
```
ol-imorozko, line 60: Do not use reference members. 
Whenever a reference is used it looks like value type (like `.` operator), but behaves like a pointer, that is can dangle, etc.
Plus now a lifetime of an `VrfLpm4Linear` object depends on a lifetime of an Allocator, but not in a safe way.

The usage of this class anyway looks like:
```cpp
//create buffer
std::unique_ptr<uint8_t[]> buffer = std::make_unique<uint8_t[]>(size_of_mem);
//create and init allocator
VrfLpm4Linear::BlocksAllocator allocator;
allocator.Init(buffer.get(), size_of_mem, YANET_RIB_VRF_MAX_NUMBER);
//create VrfLpm4Linear with allocator
VrfLpm4Linear vrf_linear(allocator);
```

So that allocator only gets used for this particular instance anyway.

Hence, you should define a move operations for `BlocksAllocator` (that's easy cause we already have a noexcept `Swap` method in it) 
```cpp
    BlocksAllocator(BlocksAllocator&& other) noexcept {
        Swap(other);
    }

    BlocksAllocator& operator=(BlocksAllocator&& other) noexcept {
        if (this != &other) {
            Swap(other);
        }
        return *this;
    }
```
and use `BlocksAllocator allocator_;` here with the constructor:

```cpp
VrfLpm4Linear::VrfLpm4Linear(BlocksAllocator&& allocator) :
        allocator_(std::move(allocator))
{
	Clear();
}
```

dataplane/vrf.h
```diff
@@ -0,0 +1,91 @@
...
+	VrfLpm4Linear(BlocksAllocator& allocator);
+
+	eResult Insert(stats_t& stats, tVrfId vrfId, const uint32_t& ipAddress, const uint8_t& mask, const uint32_t& valueId);
+	eResult Remove(stats_t& stats, tVrfId vrfId, const uint32_t& ipAddress, const uint8_t& mask);
+	eResult Clear();
+	eResult CopyFrom(const VrfLpm4Linear& other, stats_t& stats);
+	void Swap(VrfLpm4Linear& other);
+	void Lookup(const uint32_t* ipAddresses, const tVrfId* vrfIds, uint32_t* valueIds, const unsigned int& count) const;
+	std::vector<std::tuple<tVrfId, uint32_t, uint8_t, uint32_t>> GetFullList() const;
+
```
ol-imorozko, line 47: `const` method returning value should be marked `[[nodiscard]]`

dataplane/vrf.h
```diff
@@ -0,0 +1,91 @@
...
+		uint32_t valueId;
+		uint8_t mask;
+		uint8_t used;
+	};
+	static_assert(sizeof(OneBlock) <= size_of_chunk, "invalid size of OneBlock");
+
+	BlocksAllocator& allocator_;
+
+	OneBlock* Block(BlocksAllocator::Index index) const;
+};
```
ol-imorozko, line 62: `const` method returning value should be marked `[[nodiscard]]`

dataplane/vrf.h
```diff
@@ -0,0 +1,91 @@
...
```
ol-imorozko, line 156: The review comments for `VrfLpm4Linear` applies here too

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+std::string ip4str(uint32_t ipAddress)
+{
+	return common::ipv4_address_t(ipAddress).toString();
+}
+
+std::string ip6str(const uint8_t* ipAddress)
+{
+	return common::ipv6_address_t(ipAddress).toString();
+}
+
```
ol-imorozko: Unused functions

ol-imorozko: Nevermind, I saw you've removed them in the next commit

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+namespace dataplane::vrflpm
+{
+
+template<typename Allocator>
+void UpdateStats(Allocator& allocator, stats_t& stats)
```
ol-imorozko, line 9: `const Allocator&`

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+	stats.extended_chunks_size = blocks_total;
+	stats.extended_chunks_count = blocks_used;
+}
+
+/*
+ * Implementation VrfLpm4Linear
+ */
+
+VrfLpm4Linear::VrfLpm4Linear(BlocksAllocator& allocator) :
+        allocator_(allocator)
```
ol-imorozko, line 47:
```cpp
VrfLpm4Linear::VrfLpm4Linear(BlocksAllocator&& allocator) :
        allocator_(std::move(allocator))
```

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+eResult VrfLpm4Linear::Insert(stats_t& stats, tVrfId vrfId, const uint32_t& ipAddress, const uint8_t& mask, const uint32_t& valueId)
+{
+	if (vrfId >= YANET_RIB_VRF_MAX_NUMBER)
+	{
+		return eResult::invalidId;
+	}
+
+	BlocksAllocator::Index index = vrfId;
+	OneBlock* block = Block(index);
+	OneBlock* first_unused = nullptr;
```
ol-imorozko, line 61: IMHO `GetBlock` is more appropriate than just `Block`

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	block->ipAddress = ipAddress;
+	block->mask = mask;
+	block->valueId = valueId;
+	block->used = 1;
+	UpdateStats(allocator_, stats);
+
+	return eResult::success;
+}
```
ol-imorozko, line 106: Why not to use an unordered map from a `tuple <ipAddress, mask>` into `OneBlock*`?

This will allow us to avoid traversing linked list each time with `(block->ipAddress == ipAddress && block->mask == mask)` check when we need to update stats.

Like we can just `.find` in that map, and if it exists, call `UpdateStats` and return `eResult::success`.
If it isn't, then we can traverse the list, call `Allocate()` and cache the pointer to that block in a map for faster access next time

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+	while (block != nullptr)
+	{
+		if (block->ipAddress == ipAddress && block->mask == mask)
+		{
+			block->used = 0;
+			UpdateStats(allocator_, stats);
+			return eResult::success;
+		}
+		block = Block(block->next_block);
+	}
```
ol-imorozko, line 125: here `unordered_map` will work too, by just finding the `block` pointer in `O(1)`

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+VrfLpm4Linear::VrfLpm4Linear(BlocksAllocator& allocator) :
+        allocator_(allocator)
+{
+	Clear();
+}
+
+eResult VrfLpm4Linear::Insert(stats_t& stats, tVrfId vrfId, const uint32_t& ipAddress, const uint8_t& mask, const uint32_t& valueId)
+{
+	if (vrfId >= YANET_RIB_VRF_MAX_NUMBER)
```
ol-imorozko, line 54: use `extended_chunks_size_min` instead of `YANET_RIB_VRF_MAX_NUMBER`, we done
`constexpr static uint64_t extended_chunks_size_min = YANET_RIB_VRF_MAX_NUMBER;` for a reason

And in 13 other places in this commit where `YANET_RIB_VRF_MAX_NUMBER` gets used

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+		}
+	}
+
+	return result;
+}
+
+VrfLpm4Linear::OneBlock* VrfLpm4Linear::Block(BlocksAllocator::Index index) const
+{
+	return reinterpret_cast<VrfLpm4Linear::OneBlock*>((index == BlocksAllocator::null_block ? nullptr : allocator_[index]));
+}
```
ol-imorozko, line 227: IMHO this is cleaner:
```cpp
VrfLpm4Linear::OneBlock* VrfLpm4Linear::Block(BlocksAllocator::Index index) const
{
	if (index == BlocksAllocator::null_block)
	{
		return nullptr;
	}

	return reinterpret_cast<OneBlock*>(allocator_[index]);
}
```

dataplane/vrf.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+VrfLpm6Linear::VrfLpm6Linear(BlocksAllocator& allocator) :
+        allocator_(allocator)
+{
+	Clear();
+}
+
+eResult VrfLpm6Linear::Insert(stats_t& stats, tVrfId vrfId, const std::array<uint8_t, 16>& ipAddress, const uint8_t& mask, const uint32_t& valueId)
+{
+	if (vrfId >= YANET_RIB_VRF_MAX_NUMBER)
```
ol-imorozko, line 607: Same review comment as for `VrfLpm4Linear` applies for `VrfLpm6Linear` too

ol-imorozko (changes requested):

dataplane/unittest/meson.build
```diff
@@ -9,7 +9,11 @@ dependencies += dependency('gtest_main')
 sources = files('unittest.cpp',
                 'ip_address.cpp',
                 'hashtable.cpp',
-                'sdp.cpp')
+                'sdp.cpp',
+                'random_prefixes.cpp',
+                'vrf_lpm_linear.cpp',
+                '../vrf.cpp',
+                )
```
ol-imorozko: This should be done like in `controlplane/unittest/meson.build`, with sources defined in a separate variable:
```meson
dataplane_sources = files('../vrf.cpp',
                         )

sources = files('unittest.cpp',
                'ip_address.cpp',
                'hashtable.cpp',
                'sdp.cpp',
                'random_prefixes.cpp',
                'vrf_lpm_linear.cpp',
                'vrf_lpm_binary.cpp',
                )

arch = 'corei7'
cpp_args_append = ['-march=' + arch]

unittest = executable('yanet-dataplane-unittest',
                      [dataplane_sources, sources],
```

dataplane/unittest/random_prefixes.h
```diff
@@ -0,0 +1,19 @@
...
+{
+	unsigned max_number_node;
+	double depth_lambda;
+	double probability_create_node;
+	bool include_root;
+	unsigned rand_init_value;
+	tVrfId vrf_number;
+
+	std::string Description() const;
+};
```
ol-imorozko, line 15: `[[nodiscard]]`

dataplane/unittest/random_prefixes.cpp
```diff
@@ -0,0 +1,171 @@
...
+	std::shared_ptr<Node> left;
+	std::shared_ptr<Node> right;
+};
+
+std::shared_ptr<Node> BuildRandomTree(unsigned max_depth, const OneTestData& test_data, std::mt19937& generator)
+{
+	std::poisson_distribution<> poisson(test_data.depth_lambda);
+	std::uniform_real_distribution<> uniform(0.0, 1.0);
+
+	std::shared_ptr<Node> root = std::make_shared<Node>();
```
ol-imorozko, line 31: `auto root = std::make_shared<Node>();`

dataplane/unittest/random_prefixes.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+std::vector<ipv4_prefix_t> CreatePrefixesIpv4(const OneTestData& test_data)
+{
+	std::mt19937 generator(test_data.rand_init_value);
+	std::uniform_int_distribution<> digits_distribution(0, 1);
+
+	std::shared_ptr<Node> root = BuildRandomTree(32, test_data, generator);
+	std::queue<std::pair<std::shared_ptr<Node>, uint32_t>> work;
+	work.push({root, 0});
+	std::vector<ipv4_prefix_t> prefixes;
```
ol-imorozko, line 95: use `emplace` instead of push with a temporary object:
```cpp
	work.emplace(root, 0);
```

dataplane/unittest/random_prefixes.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+	while (!work.empty())
+	{
+		auto [node, value] = work.front();
+		work.pop();
+		if (node->has_value)
+		{
+			ipv4_address_t addr = ipv4_address_t::convert(common::ipv4_address_t(value));
+			prefixes.push_back({addr, static_cast<uint8_t>(node->depth)});
+		}
```
ol-imorozko, line 105: `prefixes.emplace_back(addr, static_cast<uint8_t>(node->depth));`

dataplane/unittest/random_prefixes.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+		if (node->left != nullptr)
+		{
+			work.push({node->left, value | CreatePartAddressIpv4(node->depth, node->left->depth, 0, digits_distribution, generator)});
```
ol-imorozko, line 109:
```cpp
auto new_value = value | CreatePartAddressIpv4(node->depth, node->left->depth, 0, digits_distribution, generator);
work.emplace(node->left, new_value);
```

dataplane/unittest/random_prefixes.cpp
```diff
@@ -0,0 +1,171 @@
...
+		}
+		if (node->right != nullptr)
+		{
+			work.push({node->right, value | CreatePartAddressIpv4(node->depth, node->right->depth, 1, digits_distribution, generator)});
```
ol-imorozko, line 113: same here

dataplane/unittest/random_prefixes.cpp
```diff
@@ -0,0 +1,171 @@
...
+		if (node->right != nullptr)
+		{
+			work.push({node->right, CreatePartAddressIpv6(value, node->depth, node->right->depth, 1, digits_distribution, generator)});
+		}
+	}
+
+	std::shuffle(prefixes.begin(), prefixes.end(), generator);
+
+	return prefixes;
+}
```
ol-imorozko, line 171: Same comments as for v4 version

dataplane/unittest/vrf_lpm_linear.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+TEST(VrfLpm4, SimpleLinear)
+{
+	std::unique_ptr<uint8_t[]> buffer = std::make_unique<uint8_t[]>(size_of_mem);
+	VrfLpm4Linear::BlocksAllocator allocator;
+	allocator.Init(buffer.get(), size_of_mem, YANET_RIB_VRF_MAX_NUMBER);
+	VrfLpm4Linear vrf_linear(allocator);
+
+	SimpleTestVrfLpm4(vrf_linear);
+}
```
ol-imorozko, line 27: We could **greatly** reduce code duplication by utilizing GoogleTests fixtures and `TYPED_TEST_SUITE` to parametrize tests for both `VrfLpmMap` and `VrfLpmLinear`

dataplane/unittest/vrf_lpm_common.h
```diff
@@ -0,0 +1,363 @@
...
+
+inline ipv6_address_t LastAddress(const ipv6_prefix_t& prefix)
+{
+	if (prefix.mask == 64)
+	{
+		return prefix.address;
+	}
+
+	uint64_t hi = rte_cpu_to_be_64(*(reinterpret_cast<const uint64_t*>(prefix.address.bytes)));
+	uint64_t low = rte_cpu_to_be_64(*(reinterpret_cast<const uint64_t*>(prefix.address.bytes + 8)));
```
ol-imorozko, line 128: use memcpy when dealing with simple u64 types instead of dereferencing a pointer with reintepret casts, much cleaner this way:
```cpp
        uint64_t hi = 0, low = 0;

        std::memcpy(&hi, prefix.address.bytes, sizeof(hi));
        std::memcpy(&low, prefix.address.bytes + 8, sizeof(low));
        
        hi = rte_cpu_to_be_64(hi);
        low = rte_cpu_to_be_64(low);
```

dataplane/unittest/vrf_lpm_common.h
```diff
@@ -0,0 +1,363 @@
...
+		low = static_cast<uint64_t>(-1);
+	}
+	else
+	{
+		low += (1ull << (128 - prefix.mask)) - 1;
+	}
+
+	ipv6_address_t result;
+	*(reinterpret_cast<uint64_t*>(result.bytes)) = rte_cpu_to_be_64(hi);
+	*(reinterpret_cast<uint64_t*>(result.bytes + 8)) = rte_cpu_to_be_64(low);
```
ol-imorozko, line 146:
```cpp
        hi = rte_cpu_to_be_64(hi);
        low = rte_cpu_to_be_64(low);
        std::memcpy(result.bytes.data(), &hi, sizeof(hi));
        std::memcpy(result.bytes.data() + 8, &low, sizeof(low));
 ```

File: dataplane/unittest/vrf_lpm_common.h [hunk 39]
```diff
@@ -0,0 +1,363 @@
...
+
+inline std::string AddressStr(const ipv4_address_t& address)
+{
+	return common::ipv4_address_t(address.address).toString();
+}
+
+inline std::string AddressStr(const uint32_t& address)
+{
+	return common::ipv4_address_t(address).toString();
+}
```
ol-imorozko, line 38: Those are functions that are used only for this .h file, so they should be `static` instead of `inline`.

ol-imorozko (changes requested):

dataplane/vrf.cpp
```diff
@@ -210,6 +226,372 @@ VrfLpm4Linear::OneBlock* VrfLpm4Linear::Block(BlocksAllocator::Index index) cons
...
+}
+
+eResult VrfLpm4BinaryTree::Insert(stats_t& stats, tVrfId vrfId, const uint32_t& ipAddress, const uint8_t& mask, const uint32_t& valueId)
+{
+	if (mask > 32 || valueId & 0xFF000000)
+	{
+		YANET_LOG_DEBUG("invalid prefix or value\n");
+		return eResult::invalidArguments;
+	}
+	if (vrfId >= YANET_RIB_VRF_MAX_NUMBER)
```
ol-imorozko, line 246: Fix the same issues as with Linear, please.
Like avoidance of reference field member, `emplace` instead of push with braces initialization: `push({...})`, should throw in constructor if the size allocated is not enough for `YANET_RIB_VRF_MAX_NUMBER`, otherwise loop like this
```cpp
	for (tVrfId vrfId = 0; vrfId < YANET_RIB_VRF_MAX_NUMBER; vrfId++)
	{
		OneBlock* block = Block(vrfId);
```
will segfault on `Block` when we'll call `allocator_[index]`

And also an observation that we can cache the blocks by a pair ipAddress + mask to avoid looking them up each time

ol-imorozko (changes requested):

dataplane/updater.h
```diff
@@ -669,4 +670,118 @@ class updater_array
...
+			return eResult::invalidId;
+		}
+		if (updaters_[vrf] == nullptr)
+		{
+			std::string name = std::string(name_) + ".vrf" + std::to_string(vrf);
+			updaters_[vrf] = std::make_unique<UpdaterType>(name.c_str(), memory_manager_, socket_id_);
+			if (updaters_[vrf] == nullptr)
+			{
+				return eResult::errorAllocatingMemory;
+			}
```
ol-imorozko, line 710: This won't work. `std::make_unique` **never** returns `nullptr` - either it succeeds, or it throws `std::bad_alloc`.

If we want to return an appropriate `eResult`, we need to use one of the `new` operator overloads that takes `std::nothrow` object:
```cpp
auto new_updater = std::unique_ptr<UpdaterType>(
    new(std::nothrow) UpdaterType(uniqueName.c_str(), memory_manager_, socket_id_)
);
```

Or you can define helper function in the `util` namespace:
```cpp
template<typename T, typename... Args>
std::unique_ptr<T> make_unique_nothrow(Args&&... args) {
    return std::unique_ptr<T>(new(std::nothrow) T(std::forward<Args>(args)...));
}
```

dataplane/updater.h
```diff
@@ -669,4 +670,118 @@ class updater_array
...
+
+	void limits(common::idp::limits::response& limits) const
+	{
+		for (size_t index = 0; index < YANET_RIB_VRF_MAX_NUMBER; index++)
+		{
+			if (updaters_[index])
+			{
+				updaters_[index]->limits(limits);
+			}
+		}
```
ol-imorozko, line 744:
```suggestion
		for (const auto& updater : updaters_)
		{
			if (updater)
			{
				updater->limits(limits);
			}
		}
```

dataplane/updater.h
```diff
@@ -669,4 +670,118 @@ class updater_array
...
+
+	void report(nlohmann::json& report) const
+	{
+		for (size_t index = 0; index < YANET_RIB_VRF_MAX_NUMBER; index++)
+		{
+			if (updaters_[index])
+			{
+				updaters_[index]->report(report);
+			}
+		}
```
ol-imorozko, line 755:
```suggestion
		for (const auto& updater : updaters_)
		{
			if (updater)
			{
				updater->report(report);
			}
		}
```

dataplane/updater.h
```diff
@@ -669,4 +670,118 @@ class updater_array
...
+	void clear()
+	{
+		for (size_t index = 0; index < YANET_RIB_VRF_MAX_NUMBER; index++)
+		{
+			if (updaters_[index])
+			{
+				updaters_[index]->clear();
+				updaters_[index] = nullptr;
+			}
+		}
```
ol-imorozko, line 767:
```suggestion
		for (const auto& updater : updaters_)
		{
			if (updater)
			{
				updater->clear();
				updater.reset();
			}
		}
```

dataplane/vrf.h
```diff
@@ -0,0 +1,40 @@
...
+			tVrfId vrf = vrfIds[index];
+			if ((vrf < YANET_RIB_VRF_MAX_NUMBER) && (lpms_[vrf] != nullptr))
+			{
+				lpms_[vrf]->lookup(ipAddresses + index, valueIds + index, 1);
+			}
+		}
+	}
+
+private:
+	std::array<InnerLpmType*, YANET_RIB_VRF_MAX_NUMBER> lpms_;
```
ol-imorozko, line 37: `std::array<InnerLpmType*, YANET_RIB_VRF_MAX_NUMBER>` is a common type, as I can see. I suggest to add an alias for it.
We will be able to utilize it `GetLpms` method of `updater_vrf_lpm` class too

ol-imorozko (changes requested):

dataplane/vrf.h
```diff
@@ -0,0 +1,40 @@
...
+		for (unsigned int index = 0; index < count; index++)
+		{
+			valueIds[index] = lpmValueIdInvalid;
+			tVrfId vrf = vrfIds[index];
+			if ((vrf < YANET_RIB_VRF_MAX_NUMBER) && (lpms_[vrf] != nullptr))
+			{
+				lpms_[vrf]->lookup(ipAddresses + index, valueIds + index, 1);
+			}
+		}
+	}
```
ol-imorozko, line 34: The correct approach would be to use iterators to `std::array` instead of pointer arithmetics, but we're so elbow deep in c-style arrays and pointer arithmetics in `worker.h` and in underlying `lpm` that such change would seem unnecessary.
Let's hope that in YANET2 we will be able to make things right (:

(that's not an issue, ignore it)

dataplane/worker.h
```diff
@@ -51,6 +51,11 @@ class tStack
 		mbufsCount = 0;
 	}
 
+	inline void copy_from(tStack& other)
```
ol-imorozko, line 54:
```suggestion
	void copy_from(tStack& other)
```
`inline` is redundant.

> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function

[link](https://en.cppreference.com/w/cpp/language/inline)

ol-imorozko (changes requested):

ol-imorozko:
I see that you pushed [Fix based on review comments](https://github.com/yanet-platform/yanet/pull/242/commits/a2e78ef19b4da3c78be4163bee8b42747c7fe6a1) commit.
Could you, please, also mark related conversations as "resolved"? Thanks :)

ol-imorozko (approved):

PR #255, opened 2024-10-07
Fixes updated slow worker mixing up packets between interfaces

TheRandomCharacter:


dataplane/base.h
```diff
@@ -22,13 +22,33 @@ class PortMapper
 	uint16_t ports_count_ = 0;
 	tPortId dpdk_ports_[std::numeric_limits<tPortId>::max() + 1]; // logical to dpdk
 	tPortId logical_ports_[std::numeric_limits<tPortId>::max() + 1]; // dpdk to logical
```
ol-imorozko, line 24: This class maps ports, but instead of using `std::unordered_map` we define two arrays with maximum size, and using `INVALID_PORT_ID` to mark absence of mapped key... Thus making this class 256 kilobytes in size? I mean, this _can_ be faster **if** we will have room for extra 256 kb in cache, which I doubt.

`std::unordered_map` is not a particularly good example, ideally we need some flat map, like `boost::flat_map` or any other open-source map implementation optimized for ints, like this, for example: https://github.com/Tessil/robin-map (MIT license)

But anyway, even an `std::unordered_map` is probably faster and certainly more convenient and readable than the current implementation

dataplane/kernel_interface_handler.cpp
```diff
@@ -96,6 +101,27 @@ KernelInterfaceWorker::KernelInterfaceWorker(const KernelInterfaceWorkerConfig&
...
+		std::swap(phy_queues_[i], other.phy_queues_[i]);
+		std::swap(forward_[i], other.forward_[i]);
+		std::swap(in_dump_[i], other.in_dump_[i]);
+		std::swap(out_dump_[i], other.out_dump_[i]);
+		std::swap(drop_dump_[i], other.drop_dump_[i]);
+	}
+	port_mapper_ = std::move(other.port_mapper_);
+	return *this;
+}
+
```
ol-imorozko, line 124: No point in defining any of this, those are `std::array`'s, they are not benefiting from moves, they will just be copied.
So delete this

ol-imorozko (changes requested):

dataplane/base.h
ol-imorozko: The unordered map from stl has the downside of being very cache-unfriendly due to the guarantee that you can obtain a pointer to an object in the map, and this pointer remains valid even after rehashing or insertion/deletion of other elements. This basically necessitates a node-based implementation with things like linked lists.

However, I can see how maps can be faster if the number of keys is significantly smaller than 65535 and if we'll use different hash map implementation which don't have this requirement. Because since the memory footprint might be much smaller than with two 64KB arrays, it could fit within an L1 cache, for example.

But of course to make a definitive statement we would need to measure the performance in our specific situation

dataplane/dpdk.cpp
```diff
@@ -0,0 +1,28 @@
+#include <dpdk.h>
+
+namespace dpdk
+{
+
+std::optional<std::string> GetNameByPort(tPortId pid)
+{
+	char cname[256];
```
ol-imorozko, line 8:
```suggestion
	std::array<char, 256> cname;
```
Wow, found new github feature, didn't know about that

dataplane/dpdk.cpp
```diff
@@ -0,0 +1,28 @@
...
+	if (int res = rte_eth_dev_get_name_by_port(pid, cname); res)
```
ol-imorozko, line 9:
```suggestion
	std::array<char, 256> cname;
	if (int res = rte_eth_dev_get_name_by_port(pid, cname.data()); res)
```
Wow, found new github feature, didn't know about that

PR #256, opened 2024-10-16
Adds cmd option to just check config

TheRandomCharacter:
Instead of starting dataplane returns result of dataplane config read and check.

### DIFF
--- a/dataplane/dataplane.cpp
+++ b/dataplane/dataplane.cpp
@@ -265,6 +265,11 @@ eResult cDataPlane::init(const std::string& binaryPath,
 	return result;
 }
 
+eResult cDataPlane::DryRun(const std::string& configFilePath)
+{
+	return parseConfig(configFilePath);
+}
+
 std::string rss_flags_to_string(uint64_t rss_flags)
 {
 	std::string flag_names;
--- a/dataplane/dataplane.h
+++ b/dataplane/dataplane.h
@@ -100,6 +100,7 @@ public:
 
 	eResult init(const std::string& binaryPath,
 	             const std::string& configFilePath);
+	eResult DryRun(const std::string& configFilePath);
 
 	void start();
 	void join();
--- a/dataplane/main.cpp
+++ b/dataplane/main.cpp
@@ -31,6 +31,7 @@ int main(int argc,
 	return 1;
 #endif
 
+	bool dry_run = false;
 	int config = argc;
 	for (int i = 1; i < argc; i++)
 	{
@@ -38,15 +39,26 @@ int main(int argc,
 		{
 			common::log::logPriority = common::log::TLOG_DEBUG;
 		}
+		else if (strcmp(argv[i], "-t") == 0)
+		{
+			dry_run = true;
+		}
 		else if (strcmp(argv[i], "-c") == 0)
 		{
 			config = i + 1;
+			++i;
 		}
 	}
 	if (config >= argc)
 	{
-		std::cout << "usage: " << argv[0] << " [-d] -c <dataplane.conf>" << std::endl;
+		std::cout << "usage: " << argv[0] << " [-d] [-t] -c <dataplane.conf>" << std::endl;
 		return 1;
+	}
+
+	if (dry_run)
+	{
+		eResult res = dataPlane.DryRun(argv[config]);
+		return (res == eResult::success) ? 0 : 2;
 	}
 
 	auto result = dataPlane.init(argv[0], argv[config]);


ol-imorozko (approved):

dataplane/dataplane.cpp
```diff
@@ -265,6 +265,11 @@ eResult cDataPlane::init(const std::string& binaryPath,
 	return result;
 }
 
+eResult cDataPlane::DryRun(const std::string& configFilePath)
+{
```
ol-imorozko, line 269:
```suggestion
eResult cDataPlane::DryRun(std::string_view configFilePath)
{
```
We can use lightweight wrapper around `const char*` instead of creating dynamic `std::string`

dataplane/dataplane.h
```diff
@@ -100,6 +100,7 @@ class cDataPlane
 
 	eResult init(const std::string& binaryPath,
 	             const std::string& configFilePath);
+	eResult DryRun(const std::string& configFilePath);
```
ol-imorozko, line 104: same here

dataplane/dataplane.cpp
```diff
@@ -265,6 +265,11 @@ eResult cDataPlane::init(const std::string& binaryPath,
...
+	return parseConfig(configFilePath);
+}
```
ol-imorozko, line 271: that means this function should also accept `std::string_view`, which in turn implies that `init` function should accept `std::string_view`.

In `parseConfig` we should do this, instead of creating `std::ifstream` from `std::string`:
```cpp
eResult cDataPlane::parseConfig(std::string_view configFilePath)
{
    eResult result = eResult::success;
    std::filesystem::path path(configFilePath);
    std::ifstream fromFileStream(path);
```

In summary, the following diff is all we need:
```diff
diff --git a/dataplane/dataplane.cpp b/dataplane/dataplane.cpp
index 97d9c20..fcf4577 100644
--- a/dataplane/dataplane.cpp
+++ b/dataplane/dataplane.cpp
@@ -74,7 +74,7 @@ cDataPlane::~cDataPlane()
 }
 
 eResult cDataPlane::init(const std::string& binaryPath,
-                         const std::string& configFilePath)
+                         std::string_view configFilePath)
 {
 	eResult result = eResult::success;
 
@@ -265,7 +265,7 @@ eResult cDataPlane::init(const std::string& binaryPath,
 	return result;
 }
 
-eResult cDataPlane::DryRun(const std::string& configFilePath)
+eResult cDataPlane::DryRun(std::string_view configFilePath)
 {
 	return parseConfig(configFilePath);
 }
@@ -1867,11 +1867,12 @@ void cDataPlane::switch_worker_base()
 	controlPlane->switchBase();
 }
 
-eResult cDataPlane::parseConfig(const std::string& configFilePath)
+eResult cDataPlane::parseConfig(std::string_view configFilePath)
 {
 	eResult result = eResult::success;
+	std::filesystem::path path(configFilePath);
+	std::ifstream fromFileStream(path);
 
-	std::ifstream fromFileStream(configFilePath);
 	if (!fromFileStream.is_open())
 	{
 		YADECAP_LOG_ERROR("can't open file '%s'\n", configFilePath.data());
diff --git a/dataplane/dataplane.h b/dataplane/dataplane.h
index f1437ee..fc737f6 100644
--- a/dataplane/dataplane.h
+++ b/dataplane/dataplane.h
@@ -99,8 +99,8 @@ public:
 	~cDataPlane();
 
 	eResult init(const std::string& binaryPath,
-	             const std::string& configFilePath);
-	eResult DryRun(const std::string& configFilePath);
+	             std::string_view configFilePath);
+	eResult DryRun(std::string_view configFilePath);
 
 	void start();
 	void join();
@@ -122,7 +122,7 @@ public:
 	std::string InterfaceNameFromPort(tPortId id) { return std::get<0>(ports[id]); };
 
 protected:
-	eResult parseConfig(const std::string& configFilePath);
+	eResult parseConfig(std::string_view configFilePath);
 	eResult parseJsonPorts(const nlohmann::json& json);
 	std::optional<std::map<tCoreId, CPlaneWorkerConfig>> parseControlPlaneWorkers(const nlohmann::json& config);
 	std::optional<std::pair<tCoreId, CPlaneWorkerConfig>> parseControlPlaneWorker(const nlohmann::json& cpwj);
```

ol-imorozko (changes requested):

GeorgyKirichenko:
I think it would be better to use `json-schema` instead, possibly somewhere outside `dataplane` or `controlpane`

PR #258, opened 2024-12-03
Consistent hashing for balancer

TheRandomCharacter:
Integrates consistent hashing library into yanet.
Adds option to create "chash" type balancer services.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "subprojects/json"]
 	path = subprojects/json
 	url = https://github.com/nlohmann/json.git
+[submodule "subprojects/chash"]
+	path = subprojects/chash
+	url = https://github.com/yanet-platform/chash.git
--- a/common/idp.h
+++ b/common/idp.h
@@ -285,6 +285,7 @@ using service = std::tuple<balancer_service_id_t, /// service id
                            tCounterId, ///< size 4
                            balancer::scheduler,
                            balancer::forwarding_method, // tunneling method (default ipip)
+                           balancer::scheduler_params,
                            uint32_t, ///< real_start
                            uint32_t, ///< real_size
                            std::optional<common::ipv4_prefix_t>, ///< ipv4_outer_source_network
--- a/common/idp.h
+++ b/common/idp.h
@@ -285,7 +285,6 @@ using service = std::tuple<balancer_service_id_t, /// service id
                            tCounterId, ///< size 4
                            balancer::scheduler,
                            balancer::forwarding_method, // tunneling method (default ipip)
-                           uint32_t, /// default_wlc_power
                            uint32_t, ///< real_start
                            uint32_t, ///< real_size
                            std::optional<common::ipv4_prefix_t>, ///< ipv4_outer_source_network
--- a/controlplane/balancer.cpp
+++ b/controlplane/balancer.cpp
@@ -811,7 +811,7 @@ void balancer_t::compile(common::idp::updateGlobalBase::request& globalbase,
 			        counter_id,
 			        scheduler,
 			        forwarding_method,
-			        balancer.default_wlc_power, // todo use scheduler_params.wlc_power when other services will be able to set it
+			        scheduler_params,
 			        (uint32_t)real_start,
 			        (uint32_t)(req_reals.size() - real_start),
 			        ipv4_outer_source_network,
@@ -1043,7 +1043,7 @@ bool balancer_t::reconfigure_wlc()
 			           effective_weight,
 			           connections] : service_reals_usage_info)
 			{
-				uint32_t wlc_power = scheduler_params.wlc_power;
+				uint32_t wlc_power = std::get<balancer::wlc_params>(scheduler_params).wlc_power;
 				if (wlc_power < 1 || wlc_power > 100)
 				{
 					wlc_power = YANET_CONFIG_BALANCER_WLC_DEFAULT_POWER;
--- a/controlplane/balancer.cpp
+++ b/controlplane/balancer.cpp
@@ -887,6 +887,12 @@ void balancer_t::flush_reals(common::idp::updateGlobalBaseBalancer::request& bal
 					}
 				}
 
+				auto reals_wlc = reals_wlc_weight.find(key);
+				if (reals_wlc != reals_wlc_weight.end() && reals_wlc->second.has_value())
+				{
+					effective_weight = reals_wlc->second.value();
+				}
+
 				uint32_t real_unordered_id = 0;
 				{
 					std::lock_guard<std::mutex> guard(reals_unordered_mutex);
@@ -930,8 +936,152 @@ void balancer_t::reconfigure_wlc_thread()
 {
 	while (!flagStop)
 	{
-		balancer_real_flush();
+		if (balancer_t::reconfigure_wlc())
+		{
+			balancer_real_flush();
+		}
 
 		std::this_thread::sleep_for(std::chrono::seconds(YANET_CONFIG_BALANCER_WLC_RECONFIGURE));
 	}
+}
+
+bool balancer_t::reconfigure_wlc()
+{
+	bool wlc_weight_changed = false;
+
+	common::idp::updateGlobalBaseBalancer::request balancer;
+	const auto balancer_real_connections = dataplane.balancer_real_connections();
+
+	std::lock_guard<std::mutex> guard(reals_enabled_mutex);
+
+	for (const auto& [module_name, balancer] : generations_config.current().config_balancers)
+	{
+
+		for (const auto& [service_id,
+		                  virtual_ip,
+		                  proto,
+		                  virtual_port,
+		                  version,
+		                  scheduler,
+		                  scheduler_params,
+		                  forwarding_method,
+		                  flags,
+		                  ipv4_outer_source_network,
+		                  ipv6_outer_source_network,
+		                  reals] : balancer.services)
+		{
+			(void)flags;
+			(void)version;
+			(void)forwarding_method;
+			(void)ipv4_outer_source_network;
+			(void)ipv6_outer_source_network;
+
+			if (scheduler != ::balancer::scheduler::wlc)
+			{
+				continue;
+			}
+
+			if (service_id >= YANET_CONFIG_BALANCER_SERVICES_SIZE)
+			{
+				continue;
+			}
+
+			std::vector<std::tuple<balancer::real_key_global_t, uint32_t, uint32_t>> service_reals_usage_info;
+			uint32_t connection_sum = 0;
+			uint32_t weight_sum = 0;
+
+			for (const auto& [real_ip, real_port, weight] : reals)
+			{
+				balancer::real_key_global_t key = {module_name, {virtual_ip, proto, virtual_port}, {real_ip, real_port}};
+				uint32_t effective_weight = weight;
+				{
+					auto it = reals_enabled.find(key);
+					if (it != reals_enabled.end())
+					{
+						if (it->second.has_value())
+						{
+							effective_weight = it->second.value();
+						}
+					}
+				}
+
+				weight_sum += effective_weight;
+
+				// don`t count connections for disabled reals - it can make other reals "feel" underloaded
+				if (effective_weight == 0)
+				{
+					continue;
+				}
+
+				common::idp::balancer_real_connections::real_key_t real_connections_key = {balancer.balancer_id,
+				                                                                           virtual_ip,
+				                                                                           proto,
+				                                                                           virtual_port.value(),
+				                                                                           real_ip,
+				                                                                           real_port.value()};
+				uint32_t connections = 0;
+				for (auto& [socket_id, real_connections] : balancer_real_connections)
+				{
+					(void)socket_id;
+
+					auto it = real_connections.find(real_connections_key);
+					if (it == real_connections.end())
+					{
+						continue;
+					}
+					connections += it->second;
+				}
+
+				connection_sum += connections;
+
+				service_reals_usage_info.emplace_back(key,
+				                                      effective_weight,
+				                                      connections);
+			}
+
+			for (auto [key,
+			           effective_weight,
+			           connections] : service_reals_usage_info)
+			{
+				uint32_t wlc_power = scheduler_params.wlc_power;
+				if (wlc_power < 1 || wlc_power > 100)
+				{
+					wlc_power = YANET_CONFIG_BALANCER_WLC_DEFAULT_POWER;
+				}
+
+				effective_weight = calculate_wlc_weight(effective_weight, connections, weight_sum, connection_sum, wlc_power);
+
+				if (reals_wlc_weight[key] != effective_weight)
+				{
+					reals_wlc_weight[key] = effective_weight;
+					real_updates.insert(key);
+					if (in_reload)
+					{
+						real_reload_updates.insert(key);
+					}
+					wlc_weight_changed = true;
+				}
+			}
+		}
+	}
+
+	return wlc_weight_changed;
+}
+
+uint32_t balancer_t::calculate_wlc_weight(uint32_t weight, uint32_t connections, uint32_t weight_sum, uint32_t connection_sum, uint32_t wlc_power)
+{
+	if (weight == 0 || weight_sum == 0 || connection_sum < weight_sum)
+	{
+		return weight;
+	}
+
+	auto wlc_ratio = std::max(1.0, wlc_power * (1 - 1.0 * connections * weight_sum / connection_sum / weight));
+	auto wlc_weight = (uint32_t)(weight * wlc_ratio);
+
+	if (wlc_weight > YANET_CONFIG_BALANCER_REAL_WEIGHT_MAX)
+	{
+		wlc_weight = YANET_CONFIG_BALANCER_REAL_WEIGHT_MAX;
+	}
+
+	return wlc_weight;
 }
--- a/controlplane/configparser.cpp
+++ b/controlplane/configparser.cpp
@@ -1670,7 +1670,20 @@ void config_parser_t::loadConfig_balancer_services(controlplane::base_t& baseNex
 			scheduler = balancer::scheduler::wlc;
 			if (exist(service_json, "scheduler_params") && exist(service_json["scheduler_params"], "wlc_power"))
 			{
-				scheduler_params.wlc_power = std::stoll(service_json["scheduler_params"]["wlc_power"].get<std::string>(), nullptr, 10);
+				auto& params = std::get<balancer::wlc_params>(scheduler_params);
+				params.wlc_power = std::stoll(service_json["scheduler_params"]["wlc_power"].get<std::string>(), nullptr, 10);
+			}
+		}
+		else if (scheduler_string == "chash")
+		{
+			scheduler = balancer::scheduler::chash;
+			if (exist(service_json, "scheduler_params"))
+			{
+				auto& params = std::get<balancer::chash_params>(scheduler_params);
+				params.segments_per_weight = service_json.value(
+				        "/scheduler_params/segments_per_weight"_json_pointer, 20);
+				params.siderings_count = service_json.value(
+				        "/scheduler_params/siderings_count"_json_pointer, 10000);
 			}
 		}
 		else
--- a/dataplane/globalbase.cpp
+++ b/dataplane/globalbase.cpp
@@ -1635,10 +1635,11 @@ inline uint64_t generation::count_real_connections(uint32_t counter_id)
 	return (sessions_created - sessions_destroyed + sessions_created_gc - sessions_destroyed_gc) / dataPlane->numaNodesInUse;
 }
 
-balancer_real_id_t* generation::evaluate_service_ring_one(balancer_real_id_t* start,
-                                                          const balancer_real_id_t* const do_not_exceed,
-                                                          const balancer_service_t& service,
-                                                          balancer_service_range_t& range)
+balancer_real_id_t* generation::evaluate_service_ring_one_wrr(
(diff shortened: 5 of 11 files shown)


controlplane/balancer.cpp
```diff
@@ -887,6 +887,12 @@ void balancer_t::flush_reals(common::idp::updateGlobalBaseBalancer::request& bal
 					}
 				}
 
+				auto reals_wlc = reals_wlc_weight.find(key);
+				if (reals_wlc != reals_wlc_weight.end() && reals_wlc->second.has_value())
+				{
+					effective_weight = reals_wlc->second.value();
+				}
```
ol-imorozko, line 894: I'll add comments that I've left at Levon's pr, since you will be merging his commit now:

We will have code duplication here with the following lines:
```cpp
auto it = reals_enabled.find(key);
if (it != reals_enabled.end())
{
	if (it->second.has_value())
	{
		effective_weight = it->second.value();
	}
}
```

Mayve we can add a private method in `balancer_t`, smth like
```cpp
 template <typename Map>
    std::optional<typename Map::mapped_type::value_type> get_effective_weight(const Map& map, const balancer::real_key_global_t& key) const {
        auto it = map.find(key);
        if (it != map.end() && it->second.has_value()) {
            return it->second.value();
        }
        return std::nullopt;
    }
```

And then use it, for example, with compound initialization:
```cpp
if (auto found_weight = get_effective_weight(reals_wlc_weight, key); found_weight) {
    effective_weight = *found_weight;
}
```

controlplane/balancer.cpp
```diff
@@ -930,8 +936,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+		                  flags,
+		                  ipv4_outer_source_network,
+		                  ipv6_outer_source_network,
+		                  reals] : balancer.services)
+		{
+			(void)flags;
+			(void)version;
+			(void)forwarding_method;
+			(void)ipv4_outer_source_network;
+			(void)ipv6_outer_source_network;
```
ol-imorozko, line 977: Recently I've merged my cleanup pr, so now we should use `YANET_GCC_BUG_UNUSED` macro instead of void cast to do this. You can check yourself:

> commit 01422b7a7fd6787dfb44288a08c17d6a74d70811
> Author: Ivan Morozko <ivan.morozko01@gmail.com>
> Date:   Wed Sep 25 18:54:52 2024 +0400
> 
>     Mark unused structured bindings with YANET_GCC_BUG_UNUSED instead of void cast
>     
>     GCC 7.5 version which we have to support for now due to Ubuntu 18 have a bug
>     (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81767) where unused variables in
>     structured bindings are incorrectly flagged as warnings, even though they
>     "can't" be used, cause they are not a real variables.
>     
>     Current code uses `(void)var;` to suppress these warnings, but this approach
>     is ugly and harder to grep.
>     
>     This commit removes `(void)var;` statements and replaces it with `YANET_GCC_BUG_UNUSED`
>     
>     Once we move to a newer GCC version where this bug is fixed,
>     we can easily remove the that macro as it is much easier to grep and remove
>     than those void(var) statements

controlplane/balancer.cpp
```diff
@@ -930,8 +936,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+				{
+					auto it = reals_enabled.find(key);
+					if (it != reals_enabled.end())
+					{
+						if (it->second.has_value())
+						{
+							effective_weight = it->second.value();
+						}
+					}
+				}
```
ol-imorozko, line 1006:
```cpp
auto found_weight = get_effective_weight(reals_enabled, key);
uint32_t effective_weight = found_weight ? *found_weight : weight;
```

controlplane/balancer.cpp
```diff
@@ -930,8 +936,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+			if (service_id >= YANET_CONFIG_BALANCER_SERVICES_SIZE)
+			{
+				continue;
+			}
+
+			std::vector<std::tuple<balancer::real_key_global_t, uint32_t, uint32_t>> service_reals_usage_info;
+			uint32_t connection_sum = 0;
+			uint32_t weight_sum = 0;
+
+			for (const auto& [real_ip, real_port, weight] : reals)
```
ol-imorozko, line 993: mb add `service_reals_usage_info.reserve(reals.size());`. I doubt we will have many `continue` statements in the following loop

controlplane/balancer.cpp
```diff
@@ -930,8 +936,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+				common::idp::balancer_real_connections::real_key_t real_connections_key = {balancer.balancer_id,
+				                                                                           virtual_ip,
+				                                                                           proto,
+				                                                                           virtual_port.value(),
+				                                                                           real_ip,
+				                                                                           real_port.value()};
+				uint32_t connections = 0;
+				for (auto& [socket_id, real_connections] : balancer_real_connections)
+				{
+					(void)socket_id;
```
ol-imorozko, line 1025: `YANET_GCC_BUG_UNUSED`

controlplane/balancer.cpp
```diff
@@ -930,8 +936,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+
+					auto it = real_connections.find(real_connections_key);
+					if (it == real_connections.end())
+					{
+						continue;
+					}
+					connections += it->second;
```
ol-imorozko, line 1032: just
```cpp
                    auto it = real_connections_map.find(real_connections_key);
                    if (it != real_connections_map.end())
                    {
                        connections += it->second;
                    }
```

controlplane/balancer.cpp
```diff
@@ -930,8 +936,152 @@ void balancer_t::reconfigure_wlc_thread()
...
+
+uint32_t balancer_t::calculate_wlc_weight(uint32_t weight, uint32_t connections, uint32_t weight_sum, uint32_t connection_sum, uint32_t wlc_power)
+{
+	if (weight == 0 || weight_sum == 0 || connection_sum < weight_sum)
+	{
+		return weight;
+	}
+
+	auto wlc_ratio = std::max(1.0, wlc_power * (1 - 1.0 * connections * weight_sum / connection_sum / weight));
+	auto wlc_weight = (uint32_t)(weight * wlc_ratio);
```
ol-imorozko, line 1079: that's obscure. I suggest:
```cpp
    auto scaled_connections = static_cast<double>(connections * weight_sum);
    auto scaled_weight = static_cast<double>(connection_sum * weight);
    double connection_ratio = scaled_connections / scaled_weight;

    double wlc_ratio = std::max(min_ratio, wlc_power * (1.0 - connection_ratio));

    auto new_weight = static_cast<uint32_t>(std::round(weight * wlc_ratio));
```

ol-imorozko (changes requested):

dataplane/globalbase.cpp
```diff
@@ -1672,42 +1672,93 @@ balancer_real_id_t* generation::evaluate_service_ring_one_wrr(
...
+std::vector<std::pair<balancer_real_id_t, decltype(balancer_real_state_t::weight)>>
+generation::ServiceWeights(const balancer_service_t* service)
+{
+	std::vector<std::pair<balancer_real_id_t, decltype(balancer_real_state_t::weight)>> weights;
+	for (uint32_t real_idx = service->real_start;
+	     real_idx < service->real_start + service->real_size;
+	     ++real_idx)
+	{
+		uint32_t real_id = balancer_service_reals[real_idx];
+		uint32_t weight = (balancer_real_states + real_id)->weight;
```
ol-imorozko, line 1684: just `balancer_real_states[real_id].weight`?

dataplane/globalbase.h
```diff
@@ -186,19 +186,36 @@ class generation
...
-	                                   balancer_service_range_t& range);
-	void evaluate_service_ring();
+	enum class ServiceRingOp
+	{
+		Update,
+		Rebuild
+	};
+
+	std::vector<std::pair<balancer_real_id_t, decltype(balancer_real_state_t::weight)>>
+	generation::ServiceWeights(const balancer_service_t* service);
```
ol-imorozko: don't need `generation::` here

dataplane/globalbase.h
```diff
@@ -186,19 +186,36 @@ class generation
...
-	                                   const balancer_service_t& service,
-	                                   balancer_service_range_t& range);
-	void evaluate_service_ring();
+	enum class ServiceRingOp
+	{
+		Update,
+		Rebuild
+	};
+
+	std::vector<std::pair<balancer_real_id_t, decltype(balancer_real_state_t::weight)>>
```
ol-imorozko, line 197: We use that long type at least three times, here, and two times in definition.
I suggest to add a `using` alias, will make it more readable.

dataplane/globalbase.cpp
```diff
@@ -1635,41 +1635,60 @@ inline uint64_t generation::count_real_connections(uint32_t counter_id)
...
-	for (uint32_t service_idx = 0; service_idx < balancer_services_count; ++service_idx)
+	for (uint32_t real_idx = service.real_start;
+	     real_idx < service.real_start + service.real_size;
+	     ++real_idx)
 	{
-		balancer_service_t* service = balancer_services + balancer_active_services[service_idx];
-
-		balancer_service_range_t* range = ring->ranges + balancer_active_services[service_idx];
+		uint32_t real_id = balancer_service_reals[real_idx];
+		balancer_real_state_t* state = balancer_real_states + real_id;
```
ol-imorozko, line 1649: `balancer_real_state_t* state = &balancer_real_states[real_id];` is **much** better in my opinion.
It immediately conveys the meaning that it's the address of an object in an array, rather than adding a uint32_t value to a pointer (which we don't even know unless we check the definition of balancer_real_states)

dataplane/globalbase.cpp
```diff
@@ -1635,41 +1635,60 @@ inline uint64_t generation::count_real_connections(uint32_t counter_id)
...
-	uint32_t weight_pos = 0;
-	for (uint32_t service_idx = 0; service_idx < balancer_services_count; ++service_idx)
+	for (uint32_t real_idx = service.real_start;
+	     real_idx < service.real_start + service.real_size;
+	     ++real_idx)
 	{
-		balancer_service_t* service = balancer_services + balancer_active_services[service_idx];
-
-		balancer_service_range_t* range = ring->ranges + balancer_active_services[service_idx];
+		uint32_t real_id = balancer_service_reals[real_idx];
```
ol-imorozko, line 1648: this should be `balancer_real_id_t` instead of u32

dataplane/globalbase.cpp
```diff
@@ -1635,41 +1635,60 @@ inline uint64_t generation::count_real_connections(uint32_t counter_id)
...
+
+void generation::evaluate_service_ring()
+{
+	balancer_service_ring_t* ring = &balancer_service_ring;
+	balancer_real_id_t* service_start = ring->reals;
+	for (uint32_t service_idx = 0;
+	     service_idx < balancer_services_count;
+	     ++service_idx, service_start += YANET_CONFIG_BALANCER_REAL_WEIGHT_MAX)
+	{
+		balancer_service_t* service = balancer_services + balancer_active_services[service_idx];
```
ol-imorozko, line 1778: same thing here, please, take an address of an object in the array

dataplane/globalbase.cpp
```diff
@@ -1635,41 +1635,60 @@ inline uint64_t generation::count_real_connections(uint32_t counter_id)
...
 
-		YADECAP_MEMORY_BARRIER_COMPILE;
+		balancer_service_range_t* range = ring->ranges + balancer_active_services[service_idx];
```
ol-imorozko, line 1781: same thing here, please, take an address of an object in the array

common/idp.h
```diff
@@ -285,6 +285,7 @@ using service = std::tuple<balancer_service_id_t, /// service id
                            tCounterId, ///< size 4
```
ol-imorozko, line 285: The commit message is... well... Works. Maybe. :)
I think you should squash some commits together to make reviewing easier. For example, start with a commit that refactors the old code, followed by a commit that adds chash. Make sure all corrections are included in the respective commits -- don’t leave fixes for later commits. Instead, squash those fixes so that each commit is complete and correct.

controlplane/configparser.cpp
```diff
@@ -1670,7 +1670,20 @@ void config_parser_t::loadConfig_balancer_services(controlplane::base_t& baseNex
 			scheduler = balancer::scheduler::wlc;
 			if (exist(service_json, "scheduler_params") && exist(service_json["scheduler_params"], "wlc_power"))
 			{
-				scheduler_params.wlc_power = std::stoll(service_json["scheduler_params"]["wlc_power"].get<std::string>(), nullptr, 10);
+				auto& params = std::get<balancer::wlc_params>(scheduler_params);
+				params.wlc_power = std::stoll(service_json["scheduler_params"]["wlc_power"].get<std::string>(), nullptr, 10);
```
ol-imorozko, line 1674: we can use implicit conversion here. `nlohmann::json` has an overloaded conversion operator for `std::string`. Since `std::stoll` expects `std::string`, we can just do `service_json["scheduler_params"]["wlc_power"]`

controlplane/configparser.cpp
```diff
@@ -1670,7 +1670,20 @@ void config_parser_t::loadConfig_balancer_services(controlplane::base_t& baseNex
...
+		else if (scheduler_string == "chash")
+		{
+			scheduler = balancer::scheduler::chash;
+			if (exist(service_json, "scheduler_params"))
+			{
+				auto& params = std::get<balancer::chash_params>(scheduler_params);
+				params.segments_per_weight = service_json.value(
+				        "/scheduler_params/segments_per_weight"_json_pointer, 20);
+				params.siderings_count = service_json.value(
+				        "/scheduler_params/siderings_count"_json_pointer, 10000);
```
ol-imorozko, line 1686: Why do we need json pointers here if we can get a `scheduler_params` in easy way? I understand that it's convenient to use them when accessing different nested fields, but here we can just do 
```cpp
    const auto& scheduler_params_json = service_json["scheduler_params"];
    auto& params = std::get<balancer::chash_params>(scheduler_params);

    params.segments_per_weight = scheduler_params_json.value("segments_per_weight", 20);
    params.siderings_count = scheduler_params_json.value("siderings_count", 10000);
```
instead

dataplane/globalbase.cpp
```diff
@@ -1685,6 +1685,7 @@ generation::ServiceWeights(const balancer_service_t* service)
 
 		weights.emplace_back(real_id, weight);
 	}
+	return weights;
 }
```
ol-imorozko: yea, like this, this should be just fixup-ed with the commit that adds `generation::ServiceWeights`

dataplane/globalbase.cpp
```diff
@@ -1702,10 +1703,18 @@ balancer_real_id_t* generation::rebuild_service_ring_one_chash(
...
-	auto updater = chash::WeightUpdater::MakeWeightUpdater(reals, service->details->segments_per_real);
-	updater.InitLookup(ServiceWeights(service), start);
+	auto updater = chash::WeightUpdater::MakeWeightUpdater(
+	        reals,
+	        service->details->siderings_count,
+	        service->details->segments_per_weight);
+	if (!updater)
+	{
+		YANET_LOG_ERROR("Failed to intialize updater for balancer service");
+		std::abort();
```
ol-imorozko, line 1713: here we can use `YANET_THROW`. We have exceptions disabled, so it will expand to `YANET_LOG_ERROR` + `std::abort()` anyway:
```cpp
#if __cpp_exceptions
#define YANET_THROW(message) throw std::runtime_error(std::string(message))
#else // __cpp_exceptions
#define YANET_THROW(message)                                               \
	do                                                                 \
	{                                                                  \
		YANET_LOG_ERROR("%s\n", std::string_view(message).data()); \
		std::abort();                                              \
	} while (0)
#endif // __cpp_exceptions
```

dataplane/type.h
```diff
@@ -208,6 +208,11 @@ struct ipv6_address_t
 		memset(bytes, 0, std::size(bytes));
 	}
 
+	bool operator<(ipv6_address_t other) const
+	{
+		return std::memcmp(bytes, other.bytes, 16) < 0;
+	}
+
```
ol-imorozko, line 215: why is that operator needed?

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "subprojects/json"]
 	path = subprojects/json
 	url = https://github.com/nlohmann/json.git
+[submodule "subprojects/chash"]
+	path = subprojects/chash
+	url = https://github.com/yanet-platform/chash.git
```
ol-imorozko, line 9: This commit should be one of the first ones

dataplane/meson.build
```diff
@@ -1,7 +1,7 @@
 dependencies = []
 dependencies += libdpdk.get_variable('dpdk_dep')
 dependencies += libjson.get_variable('nlohmann_json_dep')
-dependencies += libchash.get_variable('balancer_dep')
+dependencies += libchash.get_variable('chash_dep')
 dependencies += dependency('libsystemd')
```
ol-imorozko, line 5: Fixup that commit so that the commit introducing the build error no longer causes it

ol-imorozko (changes requested):

dataplane/globalbase.cpp
```diff
@@ -1635,41 +1635,60 @@ inline uint64_t generation::count_real_connections(uint32_t counter_id)
...
+		auto weight = state->weight;
 
-			while (weight-- > 0)
+		while (weight-- > 0)
+		{
+			if (start == do_not_exceed)
 			{
-				ring->reals[weight_pos++] = real_id;
+				YANET_LOG_ERROR("Balancer service exceeded ring chunk bounds\n");
+				goto finish;
```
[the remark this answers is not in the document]
TheRandomCharacter: added deferer object

dataplane/globalbase.cpp
```diff
@@ -1612,6 +1613,7 @@ balancer_real_id_t* generation::rebuild_service_ring_one_wrr(
         const balancer_real_id_t* const do_not_exceed,
         const balancer_service_t* service)
 {
+	utils::Deferer defer([]() { YADECAP_MEMORY_BARRIER_COMPILE; });
 	for (uint32_t real_idx = service->real_start;
```
ol-imorozko, line 1617: IMHO this makes code unnecessary complex. Maybe with more complex logic it will be fine, but here I feel like just doing
```cpp
//...
			if (start == do_not_exceed)
			{
				YANET_LOG_ERROR("Balancer service exceeded ring chunk bounds\n");
				YADECAP_MEMORY_BARRIER_COMPILE
				return start;
			}
//...
	YADECAP_MEMORY_BARRIER_COMPILE;
	return start;
}
```
is sufficient enough

dataplane/type.h
TheRandomCharacter: ipv6_address_t is used as real for chash. For chash to resolve collisions consistently reals need to have some ordering. < seemed to be trivial choice.

dataplane/globalbase.h
```diff
@@ -321,6 +338,7 @@ class generation
 
 	balancer_real_state_t balancer_real_states[YANET_CONFIG_BALANCER_REALS_SIZE];
 	balancer_service_ring_t balancer_service_ring;
+	std::map<const balancer_service_t*, chash::WeightUpdater> chash_updaters;
```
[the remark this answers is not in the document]
TheRandomCharacter: preferred compactness over lookup speed since `chash_updaters` are used only during reconfiguration.

dataplane/globalbase.cpp
TheRandomCharacter: We need to `YADECAP_MEMORY_BARRIER_COMPILE` at each function exit path. Using it at each existing function exit doesn't convey this intent to future function changes.

How about using a `DEFERED_MEMORY_BARRIER_COMPILE` macro?

ol-imorozko: Yeah, maybe your solution makes more sense that I've thought initially.
smth like this will suffice?
```cpp
struct MemoryBarrierGuard {
    ~MemoryBarrierGuard() {
        YADECAP_MEMORY_BARRIER_COMPILE;
    }
};
```

And do that at the start of every method:
```cpp
Result cControlPlane::init(bool use_kernel_interface)
{
    MemoryBarrierGuard guard;
    //...
```

ol-imorozko:
LGTM apart from those three unresolved conversations:
https://github.com/yanet-platform/yanet/pull/258/files#r1867724542
https://github.com/yanet-platform/yanet/pull/258/files#r1867703973
https://github.com/yanet-platform/yanet/pull/258#discussion_r1882153363

controlplane/balancer.cpp
TheRandomCharacter: this was resolved in that that pr

controlplane/balancer.cpp
TheRandomCharacter: this is not related to chash, made another pr: https://github.com/yanet-platform/yanet/pull/282

dataplane/globalbase.cpp
```diff
@@ -1622,41 +1621,166 @@ inline uint64_t generation::count_real_connections(uint32_t counter_id)
...
+			{
+				end = update_service_ring_one_chash(start, do_not_exceed, service);
+			}
+			{
+				std::unordered_map<balancer_real_id_t, std::size_t> freq;
+				for (auto c = start; c != end; ++c)
+				{
+					freq[*c]++;
+				}
+			}
```
ol-imorozko, line 1757: I guess that was left from a debug, remove it

ol-imorozko (dismissed):

PR #264, opened 2024-12-28
Add bird in Autotest.

Markuu-s:


autotest/autotest.cpp
```diff
@@ -2006,6 +2016,42 @@ bool tAutotest::step_dumpPackets(const YAML::Node& yamlStep,
...
 }
 
+bool tAutotest::step_bird(const YAML::Node& yamlStep,
+                          const std::string& path)
+{
+	std::string confPath;
+	for (const auto& yamlIter : yamlStep)
+	{
+		confPath = yamlIter.as<std::string>();
+	}
```
ol-imorozko, line 2026: why do we reinitializing that string in a loop?
Maybe just `std::string confPath = yamlStep[0].as<std::string>();` is enough?

autotest/autotest.cpp
```diff
@@ -2006,6 +2016,42 @@ bool tAutotest::step_dumpPackets(const YAML::Node& yamlStep,
...
+
+	if (!confPath.empty())
+	{
+		const std::string execCommand = "bird -c " + path + confPath;
+
+		std::string resExec = exec(execCommand.c_str());
+
+		printf("%s", resExec.data());
```
ol-imorozko, line 2034: just
```suggestion
		std::cout <<resExec;
```

Or use `YANET_LOG_`_something_. Using plain `printf` seems strange, after all, we're using C++

autotest/autotest.cpp
```diff
@@ -2006,6 +2016,42 @@ bool tAutotest::step_dumpPackets(const YAML::Node& yamlStep,
...
+	return false;
+}
+
+bool tAutotest::step_birdc(const YAML::Node& yamlStep)
+{
+	std::string command = "birdc ";
+	for (const auto& yamlIter : yamlStep)
+	{
+		command += yamlIter.as<std::string>() + " ";
+	}
```
ol-imorozko, line 2047: Use `std::ostringstream` when building strings in a loop. Current approach is not efficient as you will gradually create more and more objects of increasing size.
```suggestion
    std::ostringstream command_stream;
    command_stream << "birdc ";

    for (const auto& yamlIter : yamlStep)
    {
        command_stream << yamlIter.as<std::string>() << ' ';
    }

    const std::string command = command_stream.str();
```

autotest/autotest.cpp
```diff
@@ -2006,6 +2016,42 @@ bool tAutotest::step_dumpPackets(const YAML::Node& yamlStep,
...
+
+	auto res_birdc = exec(command.c_str());
+
+	printf("birdc:\n %s\n", res_birdc.data());
```
ol-imorozko, line 2051:
```suggestion
        std::cout << "birdc:\n" << res_birdc << '\n';
```

autotest/units/001_one_port/077_bird/autotest.yaml
```diff
@@ -0,0 +1,8 @@
+steps:
+- bird: 
+  - bird/bird.conf
+- birdc:
+  - show route
+- sleep: 20
```
ol-imorozko, line 6: is 20 second sleep absolutely needed? I would like to see a lesser value though if it's necessary, then feel free to ignore

ol-imorozko (changes requested):

ol-imorozko:
Btw this pull request is set to merge to [bird-import-proto](https://github.com/yanet-platform/yanet/tree/bird-import-proto), not `main`. Is this expected?

PR #265, opened 2024-12-28
Add config-read and run thread_birds with arguments.

Markuu-s:


common/controlplaneconfig.h
```diff
@@ -141,6 +141,32 @@ class interface_t
...
+		stream.pop(flow);
+	}
+
+	void push(common::stream_out_t& stream) const
+	{
+		stream.push(socket);
+		stream.push(vrf);
+		stream.push(flow);
+	}
+
```
ol-imorozko, line 160: Please rebase your branch onto the latest `main`.
There __will__ be conflicts because, for example, we don't use `push/pop` boilerplate anymore and instead do this:
```cpp
class config_t
{
public:
	config_t() = default;

	SERIALIZABLE(nat64stateful_id, dscp_mark_type, dscp, ipv6_prefixes, ipv4_prefixes, announces, next_module);

public:
	nat64stateful_id_t nat64stateful_id;
	common::eDscpMarkType dscp_mark_type{common::eDscpMarkType::never};
	uint8_t dscp{};
	std::vector<common::ipv6_prefix_t> ipv6_prefixes;
	std::vector<common::ipv4_prefix_t> ipv4_prefixes;
	std::set<common::ip_prefix_t> announces;
	controlplane::state_timeout state_timeout;
	std::string next_module;
	common::globalBase::flow_t flow;
};
```
That's just an example of the existing code. Note how we use the `SERIALIZABLE` macro with the names of the fields used for serialization

ol-imorozko (changes requested):

common/controlplaneconfig.h
Markuu-s: Updated bird-import-proto and this branches.

common/controlplaneconfig.h
```diff
@@ -104,6 +105,19 @@ class interface_t
...
 };
 
+class bird_import_t
+{
+public:
+	SERIALIZABLE(socket, vrf);
+
+public:
+	inline static const std::string socketStr = "socket";
+	inline static const std::string vrfStr = "vrf";
```
ol-imorozko, line 115:
```suggestion
	static constexpr auto socketStr = "socket";
	static constexpr auto vrfStr = "vrf";
```

common/controlplaneconfig.h
```diff
@@ -104,6 +105,19 @@ class interface_t
...
+
+	std::string socket;
+	std::string vrf;
+};
```
ol-imorozko, line 119: Since everything is public, why not just 
```suggestion
struct bird_import_t
{
	std::string socket;
	std::string vrf;
	
	static constexpr auto socketStr = "socket";
	static constexpr auto vrfStr = "vrf";
	
	SERIALIZABLE(socket, vrf);
};
```

controlplane/configparser.cpp
```diff
@@ -197,6 +197,32 @@ controlplane::base_t config_parser_t::loadConfig(const std::string& rootFilePath
 	return baseNext;
 }
 
+void config_parser_t::loadConfig_route_bird(controlplane::base_t& baseNext,
```
ol-imorozko, line 200: I guess we need 
```suggestion
void config_parser_t::loadConfig_route_bird([[maybe_unused]] controlplane::base_t& baseNext,
```

controlplane/configparser.cpp
```diff
@@ -197,6 +197,32 @@ controlplane::base_t config_parser_t::loadConfig(const std::string& rootFilePath
...
+		if (exist(elemJson, BirdImport::vrfStr))
+		{
+			import.vrf = elemJson[BirdImport::vrfStr];
+		}
+
+		birdsImport.push_back(import);
+		YANET_LOG_INFO("loadConfig_route_bird: socket(%s), vrf(%s)\n",
+		               import.socket.data(),
+		               import.vrf.data());
+	}
```
ol-imorozko, line 223: interchange those lines, and now we will be able to benefit from moving `import` object, which is essentially two `std::string`'s
```suggestion
		YANET_LOG_INFO("loadConfig_route_bird: socket(%s), vrf(%s)\n",
		               import.socket.data(),
		               import.vrf.data());
		birdsImport.push_back(std::move(import));
	}
```

controlplane/controlplane.cpp
```diff
@@ -941,6 +953,14 @@ eResult cControlPlane::loadConfig(const std::string& rootFilePath,
...
 		YANET_LOG_INFO("dataplane has been updated (stage 7)\n");
+		for (auto& module : modules)
+		{
+			if (rib_t* rib = dynamic_cast<rib_t*>(module))
+			{
+				rib->bird_import_get();
+				rib->moduleStart();
+			}
+		}
 	}
```
ol-imorozko, line 964: Is every module is also a rib module? I don't quite understand.
Also, `dynamic_cast` will inevitably lead to some errors, if, for example, we decide to change class hierarchy.
Can we use other things here? Like maybe we can utilize one or two virtual functions?

controlplane/rib.cpp
```diff
@@ -869,22 +865,41 @@ void rib_t::rib_thread()
...
 		std::this_thread::sleep_for(std::chrono::milliseconds{200});
 	}
+}
 
+void rib_t::bird_import_get()
+{
+	auto route = controlPlane->getRoute();
+	for (auto& [vrf, response] : route)
+	{
+		(void)vrf;
```
ol-imorozko, line 875:
```suggestion
		YANET_GCC_BUG_UNUSED(vrf);
```

ol-imorozko (changes requested):

common/controlplaneconfig.h
Markuu-s: 1) "constexpr" doesn`t include "inline" in all compilers 
2) I thought about "constexpr" instead of "const" buf std::string doesn`t support it 
2) auto give us const char[N], I think in C++ it worst

common/controlplaneconfig.h
Markuu-s: I agree, but I followed the standard of writing classes in this file.

controlplane/configparser.cpp
Markuu-s: +

controlplane/rib.cpp
Markuu-s: Ok

controlplane/controlplane.cpp
Markuu-s: At this stage, I need to call one method that is only in rib_t (bird_import_get()), using a virtual function for all classes is somehow strange. I'll think about it, or I'm ready to discuss what's best.

controlplane/configparser.cpp
Markuu-s: Ok

ol-imorozko (changes requested):
We try to avoid "merge commits" in the history. You can simply squash that commit with "Add config-read and run thread_birds with arguments." In fact, commits like "Fix PR" are not desirable either. Commits should be self-contained, i.e., if you are improving/changing something after a review, it's better to introduce those changes within the commit itself. Here, you can just squash your four commits into one: "Add config-read and run thread_birds with arguments."


Also, could you please add a small description to commit body? It's hard to grasp what "run thread_birds with arguments" mean from commit title.

common/controlplaneconfig.h
ol-imorozko: 1. How is that relevant? I suggested removing `inline` here because it's redundant since you already have `static`. There's no need to allow multiple definitions with `inline` — `static` already means that the variable will have a single definition in the program.

2. Yeah, that's correct. Here you can use `std::string_view` for string literals, but the issue with it is that `nlohmann::json` doesn't have an overload for `operator[]` with `std::string_view`. However, it does have overloads for `std::string` and `const char*`. So you could either use the change I suggested, or go with `std::string_view` and pass `.data()` like this:
```cpp
if (exist(elemJson, BirdImport::socketStr))
{
    import.socket = elemJson[BirdImport::socketStr.data()];
}

if (exist(elemJson, BirdImport::vrfStr))
{
    import.vrf = elemJson[BirdImport::vrfStr.data()];
}
```
Either way works for me.

3. Why do you think so? IMHO using array types isn't inherently bad in C++. It's quite common and valid (as long as we don't decay them into pointers and/or perform pointer arithmetic :) ). I don't see any issues with using array types in this context.

controlplane/controlplane.cpp
ol-imorozko: `modules` is a vector of `cModule` pointers. So what you can do is add a new field to the class that will point to the required module, and here just call it's methods.
Something like an iterator to the `modules` will suffice (but be aware of invalidation). Or just a pointer to `rib_t`.
Either way it's better than what we have how.

ol-imorozko (changes requested):

common/controlplaneconfig.h
ol-imorozko: 1. How is that relevant? I suggested removing `inline` here because it's redundant since you already have `static`. There's no need to allow multiple definitions with `inline` — `static` already means that the variable will have a single definition in the program.

2. Yeah, that's correct. Here you can use `std::string_view` for string literals, but the issue with it is that `nlohmann::json` doesn't have an overload for `operator[]` with `std::string_view`. However, it does have overloads for `std::string` and `const char*`. So you could either use the change I suggested, or go with `std::string_view` and pass `.data()` like this:
```cpp
if (exist(elemJson, BirdImport::socketStr))
{
    import.socket = elemJson[BirdImport::socketStr.data()];
}

if (exist(elemJson, BirdImport::vrfStr))
{
    import.vrf = elemJson[BirdImport::vrfStr.data()];
}
```
Either way works for me.

3. Why do you think so? IMHO using array types isn't inherently bad in C++. It's quite common and valid (as long as we don't decay them into pointers and/or perform pointer arithmetic :) ). I don't see any issues with using array types in this context.

controlplane/controlplane.cpp
ol-imorozko: `modules` is a vector of `cModule` pointers. So what you can do is add a new field to the class that will point to the required module, and here just call it's methods.
Something like an iterator to the `modules` will suffice (but be aware of invalidation). Or just a pointer to `rib_t`.
Either way it's better than what we have how.

controlplane/configparser.cpp
```diff
@@ -197,6 +197,32 @@ controlplane::base_t config_parser_t::loadConfig(const std::string& rootFilePath
...
+
+		if (exist(elemJson, BirdImport::socketStr))
+		{
+			import.socket = elemJson[BirdImport::socketStr];
+		}
+
+		if (exist(elemJson, BirdImport::vrfStr))
+		{
+			import.vrf = elemJson[BirdImport::vrfStr];
+		}
```
ol-imorozko, line 217: What if those keys does not exist? We need to use some default value then, right? Otherwise I think it will be two empty strings.
It's better to use `value(const typename object_t::key_type& key, const ValueType& default_value)` method of `nlohmann::json` like so:
```suggestion
		import.socket = elemJson.value(BirdImport::socketStr, "some_default_value");
		import.vrf = elemJson.value(BirdImport::vrfStr, "some_default_value");
```

PR #269, opened 2025-01-30
Fix balancer round robin

darkgrey1554:


controlplane/configparser.cpp
```diff
@@ -1743,6 +1743,11 @@ void config_parser_t::loadConfig_balancer_services(controlplane::base_t& baseNex
 			flags |= YANET_BALANCER_OPS_FLAG;
 		}
 
+		if (service_json.value("pure_round_robin", false))
```
darkgrey1554, line 1746: maybe its worth changing the field name ("prb" or "round_robin")

ol-imorozko (dismissed):

ol-imorozko:
Could you, please, correct commit history? It's not desirable to push commits with names like "fix", "fix format" etc. to `main`.
It you fix something for a commit, make sure to squash that change with the commit itself. 
I.e here we should have only two commits: 
"Add pure round robin to balancer"
and
"Add purr scheduler"

darkgrey1554:
> [ ](/darkgrey1554)

Combined everything into one commit )

ol-imorozko (approved):

GeorgyKirichenko (approved):

PR #274, opened 2025-02-07
Support ICMP Time Exceeded Message #issues57

darkgrey1554:
YANET doesn't support ICMP response, when receiving packet with TTL = 1.
By RFC792 YANET should send ICMP Time Exceeded Message,
when packet cannot be forwarded or balanced.

The icmp response should contain the IP of the current local device. That's why we added a new config:

"hostConfig":{
	"addresses": [string],
	"showRealAddress": bool
  }

addresses - list of host machine addresses (ipv4 and ipv6)
showRealAddress - bool flag, enables transmission of the machine's IP address in the ICMP response


dataplane/worker.cpp
```diff
@@ -5970,3 +5980,135 @@ inline void cWorker::populate_hitcount_map(const std::string& id, rte_mbuf* mbuf
 	entry.count++;
 	entry.bytes += mbuf->pkt_len;
 }
+
+inline rte_mbuf* cWorker::create_icmp_package_time_exceeded(rte_mbuf* mbuf)
+{
+	const auto& host_config = bases[localBaseId & 1].globalBase->host_config;
+	rte_mbuf* new_mbuf = rte_pktmbuf_alloc(mempool);
```
ezhk, line 5987: Can we modify exist package without copy?

autotest/units/001_one_port/081_time_exceeded_in_transit/gen.py
```diff
@@ -0,0 +1,54 @@
...
+	return Ether(dst="00:11:22:33:44:55", src="00:00:00:22:22:22")/Dot1Q(vlan=200)/IPv6(src=_src, dst=_dst, hlim=64, fl=0)
+
+def ipv6_recv(_src, _dst):
+	return Ether(dst="00:00:00:11:11:11", src="00:11:22:33:44:55")/Dot1Q(vlan=100)/IPv6(src=_src, dst=_dst, hlim=63, fl=0)
+
+write_pcap("001-send.pcap",
+        Ether(dst="00:11:22:33:44:55", src="00:00:00:11:11:11")/Dot1Q(vlan=200)/IP(src="10.0.0.3", dst="10.0.0.55", ttl=64)/TCP(dport=80, sport=1024, flags=""),
+		Ether(dst="00:11:22:33:44:55", src="00:00:00:22:22:22")/Dot1Q(vlan=200)/IPv6(src="2000::1:b", dst="2000::1:bc", hlim=64, fl=0)/UDP(dport=53, sport=1024),
+		Ether(dst="00:11:22:33:44:55", src="00:00:00:11:11:11")/Dot1Q(vlan=200)/IP(src="10.0.0.3", dst="10.0.0.55", ttl=1)/TCP(dport=80, sport=1024, flags=""),
+		Ether(dst="00:11:22:33:44:55", src="00:00:00:11:11:11")/Dot1Q(vlan=100)/IPv6(src="3456::a", dst="3456::b", hlim=1, fl=0)/TCP(dport=80, sport=1024, flags=""))
```
ezhk, line 35: Lets try to add TTL values at the range boundary: ttl=0 (ICMP should be crafted), ttl=2 (is ok).

autotest/units/001_one_port/081_time_exceeded_in_transit/controlplane_full_config_hidden_ip_disabled.conf
```diff
@@ -0,0 +1,49 @@
+{
+  "host_config":{
+	"addresses":[
+		"10.0.0.10",
+		"3456::c"
+	],
+	"hidden_real_ip": false
```
ezhk, line 7: `hidden_real_ip` -> ShowRealAddress
`host_config` -> HostConfig

below in config is using camel case.

dataplane/worker.cpp
```diff
@@ -1694,6 +1695,15 @@ inline void cWorker::acl_ingress_flow(rte_mbuf* mbuf,
 	dataplane::metadata* metadata = YADECAP_METADATA(mbuf);
 	metadata->flow = flow;
 
+	if (flow != common::globalBase::eFlowType::route_local && !is_valid_ttl(mbuf))
```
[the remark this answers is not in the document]
darkgrey1554: Ok) changed the function name to "is_expired_ttl". Made the packet creation via the general function "create_icmp_package" (created ICMP packet types). For convenience, moved the "calcHash" function to metadata to  metadata.h.

dataplane/worker.cpp
darkgrey1554: We decided that it is necessary to register drops for such packages. Therefore, we send the original package to the drop for its correct destruction and registration of statistics.

autotest/units/001_one_port/081_time_exceeded_in_transit/gen.py
darkgrey1554: Added test cases for ttl=0 and ttl=2

autotest/units/001_one_port/081_time_exceeded_in_transit/controlplane_full_config_hidden_ip_disabled.conf
darkgrey1554: ok ) fixed it

dataplane/worker.cpp
```diff
@@ -5970,3 +5980,135 @@ inline void cWorker::populate_hitcount_map(const std::string& id, rte_mbuf* mbuf
 	entry.count++;
 	entry.bytes += mbuf->pkt_len;
 }
+
+inline rte_mbuf* cWorker::create_icmp_package_time_exceeded(rte_mbuf* mbuf)
```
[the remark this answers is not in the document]
darkgrey1554: done)

dataplane/icmp.cpp
```diff
@@ -0,0 +1,147 @@
+#include "icmp.h"
+
+#include "metadata.h"
+
+namespace yanet::icmp
+{
+
+bool create_icmp_package(TypePackage type, rte_mbuf* mbuf_target, const CreatePackagePayload& payload)
```
darkgrey1554, line 8: It is not necessary to specify the package type, but I think it makes it clearer.

dataplane/worker.cpp
```diff
@@ -392,62 +394,6 @@ YANET_NEVER_INLINE void cWorker::mainThread()
 	}
 }
 
-inline void cWorker::calcHash(rte_mbuf* mbuf, uint8_t flags)
```
darkgrey1554, line 395: I moved the function to metadata.h. It needed in icmp.cpp. And it seems more logical that it will be in metadata.h

ol-imorozko:
`Add support ICMP Time Exceeded Message` is grammatically incorrect. I suggest to just use `Support ICMP Time Exceeded Message`.
Also it would be really helpful if you add commit description about why we need this. You can just copy issue description and have smth like:
```
Support ICMP Time Exceeded Message

YANET doesn't support ICMP response, when receiving packet with TTL = 1.
By RFC792 YANET should send ICMP Time Exceeded Message, when packet cannot be forwarded or balanced.
```

dataplane/globalbase.cpp
```diff
@@ -424,6 +424,10 @@ eResult generation::update(const common::idp::updateGlobalBase::request& request
 		{
 			result = tscs_base_value_update(std::get<common::idp::updateGlobalBase::tscs_base_value_update::request>(data));
 		}
+		else if (type == common::idp::updateGlobalBase::requestType::update_host_config)
+		{
+			result = update_host_config(std::get<common::idp::updateGlobalBase::update_host_config::request>(data));
+		}
```
ol-imorozko, line 430: This commit is about supporting some kind of ICMP response message.
Could you clarify these changes please?
I'm not following why we have this "update_host_config" sub-theme going on.

common/type.h
```diff
@@ -526,6 +526,15 @@ class ipv6_address_t
...
 
+	[[nodiscard]] bool is_default() const
+	{
+		if (operator==(ipv6_address_t()))
+		{
+			return true;
+		}
+		return false;
+	}
+
```
ol-imorozko: It's not efficient to construct a temporary object for comparison.
I suggest
```cpp
[[nodiscard]] bool is_default() const
{
    return std::all_of(address.begin(), address.end(), [](uint8_t byte) {
        return byte == 0;
    });
}
```

dataplane/type.h
```diff
@@ -131,6 +131,13 @@ struct ipv4_address_t
 		return result;
 	}
 
+	[[nodiscard]] bool is_default() const
+	{
+		if (address == 0)
+			return true;
+		return false;
```
ol-imorozko, line 138: `return (address == 0);`

ol-imorozko (changes requested):

dataplane/type.h
darkgrey1554: fix it

dataplane/globalbase.cpp
darkgrey1554: The icmp response should contain the IP of the current local device. It was decided that the IP of the local device will be obtained from the configuration.
A new section was introduced in the config (), which is passed to Dataplane

ol-imorozko: Then it should be done in a separate commit, otherwise it's hard to grasp the intentions, especially from the commit title.
I suggest to split the first commit into two -- one that will add that update mechanism with commit description equal to what you told me here, and the other that will actually do ICMP-related things

dataplane/type.h
```diff
@@ -191,14 +196,8 @@ struct ipv6_address_t
...
-		for (auto b : bytes)
-		{
-			if (b != 0)
-			{
-				return false;
-			}
-		}
-		return true;
+		uint8_t zero_buf[16]{};
+		return !memcmp(bytes, zero_buf, std::size(bytes));
```
ol-imorozko, line 200: Still, we don't need to create temporary object/array for comparison. `std::all_of` is perfectly fine:
```cpp
[[nodiscard]] bool is_default() const
{
    return std::all_of(address.begin(), address.end(), [](uint8_t byte) {
        return byte == 0;
    });
}
```

dataplane/icmp.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+#include "metadata.h"
+
+namespace yanet::icmp
+{
+
+namespace
+{
+const uint8_t TTL = 128;
+}
```
ol-imorozko, line 11: We don't need an anonymous namespace since we're using that only in this cpp file.
I suggest just `constexpr uint8_t TTL = 128;`

dataplane/icmp.cpp
```diff
@@ -0,0 +1,156 @@
...
+			{
+				rte_memcpy(ip_header->dst_addr, ip_header->src_addr, 16);
+				rte_memcpy(ip_header->src_addr, host_config.ipv6_address.bytes, 16);
+			}
+			else
+			{
+				uint8_t addr[16];
+				rte_memcpy(addr, ip_header->dst_addr, 16);
+				rte_memcpy(ip_header->dst_addr, ip_header->src_addr, 16);
+				rte_memcpy(ip_header->src_addr, addr, 16);
```
ol-imorozko, line 93: I believe we can use `RTE_SWAP` for this

dataplane/icmp.cpp
```diff
@@ -0,0 +1,156 @@
...
+	icmp_header->icmp_seq_nb = 0;
+	icmp_header->icmp_cksum = 0;
+	if (metadata->transport_headerType == IPPROTO_ICMP)
+	{
+		icmp_header->icmp_type = 11;
+
+		// copy network header and next 64bits of source package (RFC 792)
+		rte_ipv4_hdr* ip_header = rte_pktmbuf_mtod_offset(mbuf_target, rte_ipv4_hdr*, metadata->network_headerOffset);
+		rte_memcpy(rte_pktmbuf_mtod_offset(mbuf_target, char*, metadata->transport_headerOffset + sizeof(rte_icmp_hdr)),
+		           rte_pktmbuf_mtod_offset(payload.mbuf_source, char*, YADECAP_METADATA(payload.mbuf_source)->network_headerOffset),
```
ol-imorozko, line 124: why `YADECAP_METADATA(payload.mbuf_source)` instead of just `metadata`?

dataplane/icmp.cpp
```diff
@@ -0,0 +1,156 @@
...
+		mbuf_target->pkt_len = metadata->network_headerOffset + rte_cpu_to_be_16(ip_header->total_length);
+		mbuf_target->data_len = mbuf_target->pkt_len;
+	}
+	else
+	{
+		icmp_header->icmp_type = 3;
+		rte_ipv6_hdr* ip_header = rte_pktmbuf_mtod_offset(mbuf_target, rte_ipv6_hdr*, metadata->network_headerOffset);
+
+		rte_memcpy(rte_pktmbuf_mtod_offset(mbuf_target, char*, metadata->transport_headerOffset + sizeof(rte_icmp_hdr)),
+		           rte_pktmbuf_mtod_offset(payload.mbuf_source, char*, YADECAP_METADATA(payload.mbuf_source)->network_headerOffset),
```
ol-imorozko, line 139: same here

ol-imorozko (changes requested):

dataplane/icmp.cpp
darkgrey1554: I tried, but the original data type is array and the compiler gives an error " error: array must be initialized with a brace-enclosed initializer". I would not like to make corrections in "RTE_SWAP"

dataplane/type.h
darkgrey1554: rolled back the changes

ol-imorozko: Oh, I see, you removed "is_default" for ipv6 and you use `ipv6_address.empty()` instead. That's fine my be

dataplane/globalbase.cpp
ol-imorozko: Okay, fine, but please add description to the commit message. Both:
- description taken from the issue, that will explain why we need this commit
- explanation of those host_config changes, aka what you told me here in this conversation.
Thanks!

darkgrey1554: Added information about the config to the description

GeorgyKirichenko (approved):

ol-imorozko (approved):

PR #275, opened 2025-02-11
Implemented reading in multiple threads of data from different bird tables

stal76:
Autotests and an application for testing the reading of route information from bird have also been added.


ol-imorozko:
It's better to use the imperative mood for commit messages. I suggest:

PR name:
`Implement reading data from multiple threads across different bird tables`

Commits:
```
Start bird import thread (Already good)
Add ability to read route information from multiple bird tables
Emulate working with bird in autotests
Add autotests that emulate reading the routing table from bird
Add application for testing route information reading from bird
```

ol-imorozko (changes requested):
The current approach with raw memory and `decode_*` functions is problematic and I want it changed.

Casting raw memory to different types violates [strict aliasing rules](https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8) and can lead to UB.

Instead we should use `std::istream` which is an abstraction over a character sequence, which is precisely what we're operating with here. It provides type-safe methods for reading data and automatically handles boundary checking.

This way, instead of
```cpp
static inline bool decode_u8(uintptr_t* ppos, uintptr_t end, uint8_t* pvalue)
{
    if (*ppos + sizeof(uint8_t) > end)
        return false;
    *pvalue = *(uint8_t*)*ppos;
    *ppos += sizeof(uint8_t);
    return true;
}
```
we will have just
```cpp
static inline bool decode_u8(std::istream& stream, uint8_t& value)
{
    return !!stream.read(reinterpret_cast<char*>(&value), sizeof(value));
}
```

**I will** propose specific changes to replace raw memory decoding with `std::istream` in the related places with review comments by myself, so I'm not asking you to rewrite the whole file by yourself.

controlplane/libbird.cpp
```diff
@@ -0,0 +1,729 @@
...
+ * @vrf: the string with the vrf name
+ * @handler: handler events for receiving data from bird
+ * @pipe_close: the pipe descriptor through which information about the need to terminate the stream will
+ *              be transmitted, a negative value means that verification should be disabled.
+ *
+ * return value:
+ * - true: some kind of error has occurred and it may be worth running the function after some time interval.
+ * - false: data was received through pipe indicating that the thread needs to be completed
+ */
+bool read_bird_feed(const char* sock_name, const char* vrf, rib_update_handler handler, int pipe_close)
```
ol-imorozko, line 608:
```suggestion
bool read_bird_feed(const char* sock_name, const std::string& vrf, rib_update_handler handler, int pipe_close)
```

controlplane/libbird.cpp
```diff
@@ -0,0 +1,729 @@
...
+			if (!decode_chunk(&pos, end, &route_end))
+				break;
+
+			common::icp::rib_update::action action;
+			if (!parse_route_update(&pos, route_end, vrf, &action))
+				break;
+
+			requests.emplace_back(action);
+
+			parse_pos = pos - (uintptr_t)read_buf;
```
ol-imorozko, line 705:
```suggestion
			/* Create a stringstream from the current buffer */
			std::stringstream stream;
			stream.write((char*)((uintptr_t)read_buf + read_pos), read_pos - parse_pos);

			std::streampos route_end;
			if (!decode_chunk(stream, route_end))
				break;

			common::icp::rib_update::action action;
			if (!parse_route_update(stream, vrf, action))
				break;

			requests.push_back(std::move(action));

			parse_pos += stream.tellg();
```

ol-imorozko (changes requested):

ezhk:
What's status of this PR?

PR #276, opened 2025-02-13
ACL compilation improvements for v64

ol-imorozko:
Added a new Meson option `-Dcustom_hash_structures`, which is enabled by default. When set, ACL compilation will use flatmap and flatset instead of standard maps and sets. These structures are not node-based, meaning they invalidate iterators and are not drop-in replacements for `std::unordered_map` and `std::unordered_set`. However, if iterator stability is not required, they offer significant performance benefits.  

### Benchmark Results
Tested on i7-12800H 4.7 GHz, 32GiB DDR5 4800 MHz, Ubuntu 22.04.5, GCC 11.4.0

10 runs were performed, and plots show results with a 95% confidence interval.  

time
memory

- Original: Uses the most memory but is faster than the improved version with standard containers.  
- Improved with std containers: Uses significantly less memory but is slightly slower.  
- Improved with flat containers: Best overall -- fastest execution and lowest memory usage.  

Results:
**62.17%** faster  
**41.18%** less memory used

### DIFF
--- a/controlplane/unittest/meson.build
+++ b/controlplane/unittest/meson.build
@@ -5,6 +5,10 @@ dependencies += dependency('threads')
 dependencies += dependency('gtest')
 dependencies += dependency('gtest_main')
 dependencies += libyanet_fwparser_dep
+if get_option('custom_hash_structures')
+    dependencies += emhash_dep
+    dependencies += ankerl_dep
+endif
 
 controlplane_sources = files('../acl_compiler.cpp',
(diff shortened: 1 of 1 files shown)


ol-imorozko:
I rebased this branch on `(tag: v64.0)`. Should I rebase it to `release/64` instead?
@GeorgyKirichenko

GeorgyKirichenko (approved):

PR #284, opened 2025-03-20
WIP: Refactor route tunnel compile

TheRandomCharacter:


common/utils.h
```diff
@@ -46,5 +46,15 @@ inline std::vector<std::string> split(const std::string& str, char delimiter)
...
 
+// helper type for the variant visitor
+template<class... Ts>
+struct Visitor : Ts...
+{
+	using Ts::operator()...;
+};
+// explicit deduction guide (not needed as of C++20)
+template<class... Ts>
+Visitor(Ts...) -> Visitor<Ts...>;
```
ol-imorozko, line 57: That's cool.
Personally, I would do a regular approach with `if constexpr`:
```cpp
	const auto visitor = [&](auto&& arg) {
		using T = std::decay_t<decltype(arg)>;

		if constexpr (std::is_same_v<T, route::tunnel_destination_interface_t>)
		{
		}
		else if constexpr (std::is_same_v<T, route::tunnel_destination_legacy_t>)
		{
		}
		else if constexpr (std::is_same_v<T, route::directly_connected_destination_t>)
		{
		}
		else if constexpr (std::is_same_v<T, route::tunnel_destination_default_t>)
		{
		}
	};

	std::visit(visitor, destination);
```

But I don't see any problems with your approach.

ol-imorozko (approved):

common/utils.h
ol-imorozko: apparently it's a popular pattern: https://www.cppstories.com/2019/02/2lines3featuresoverload.html/

I think we should name it `Overload`/`Overloaded` then

PR #290, opened 2025-04-14
Make common a meson dependency

TheRandomCharacter:
Currently a crutch with yanet_rootdir is used to enable meson modules cli,controlplane,dataplane,... to include common headers, that are out of source tree for them.

This change makes common a dependency for those modules to depend on.


ol-imorozko (approved):
Why do we use `common/include/common/some_file.h`?
I feel like `common/include/some_file.h` is enough

PR #310, opened 2025-08-19
Llocal

TheRandomCharacter:
Replaces hodgepodge neighbor discovery over netlink with common libnl library.

Untangles neighbor module from dataplane and netlink logic.

Covers neighbor tracking logic with unit testing.


dataplane/netlink.cpp
```diff
@@ -0,0 +1,261 @@
...
+	Entry entry;
+	char ifname[IFNAMSIZ];
+	if (if_indextoname(sysifid, ifname) == nullptr)
+	{
+		YANET_LOG_INFO("Skipping message for unknown OS interface '%i'\n", sysifid);
+		return NL_OK;
+	}
+	if (auto id = get_id(ifname); id.has_value())
+	{
+		std::get<tInterfaceId>(entry) = id.value();
```
ol-imorozko, line 26: using a structure will avoid those `std::get`'s that are all over the place right now

dataplane/netlink.cpp
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	return entry;
+}
+
+std::vector<Entry> Provider::GetHostDump(
+        const std::unordered_map<std::string, tInterfaceId>& ids)
+{
+	rtgenmsg rt_hdr = {.rtgen_family = AF_UNSPEC};
+	nl_sock* sk = nl_socket_alloc();
```
ol-imorozko, line 75: (see review comment to netlink.hpp)

we will avoid using goto:
```suggestion
	std::unique_ptr<nl_sock, nl_sock_deleter> sk(nl_socket_alloc());
```

dataplane/unittest/neighbor.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include <iostream>
+
+#include "../neighbor.h"
+
+namespace
+{
+
+class MockProvider : public netlink::Interface
+{
+public:
```
ol-imorozko, line 12: You can just mark the whole class as `final` instead of each method

dataplane/dataplane.cpp
```diff
@@ -286,7 +286,48 @@ eResult cDataPlane::init(const std::string& binaryPath,
...
+					        auto key = *iter.key();
+
+					        iter.unset_valid();
+					        iter.unlock();
+
+					        keys.emplace_back(key);
+				        }
+
+				        return true;
+			        });
```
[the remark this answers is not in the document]
TheRandomCharacter: Nice catch

ol-imorozko (approved):

GeorgyKirichenko (approved):

PR #332, opened 2026-07-23
Add source_ipv6_prefixes parametr to nat64stateful

glumpo:
Adds optional `source_ipv6_prefixes` field support to nat64stateful modules.

The control plane now:

- Selects a NAT64 module using both source and destination IPv6 prefixes.
- Preserves existing behavior with `::/0` as the default source prefix.
- Exposes source prefixes in the `nat64stateful` CLI output.

### DIFF
--- a/autotest/units/001_one_port/061_nat64stateful/autotest.yaml
+++ b/autotest/units/001_one_port/061_nat64stateful/autotest.yaml
@@ -1,6 +1,11 @@
 steps:
 - ipv4Update: "0.0.0.0/0 -> 200.0.0.1"
 - ipv6Update: "::/0 -> fe80::1"
+- cli_check: |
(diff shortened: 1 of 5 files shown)


ezhk:
There is a source-prefix parsing issue to fix before merge.

`source_ipv6_prefixes` validates the mask with `std::stoul(mask_string)` (decimal), but then constructs `ipv6_prefix_t` from the original string. That constructor parses the mask with base `0`, so leading-zero masks are interpreted as octal: for example, `2001:db8::/064` passes validation but becomes `/52`, while `::/08` becomes `/0`. This can make the generated ACL select the NAT64 module for a much broader source range than configured.

Please construct the prefix using the already-parsed decimal mask (or reject noncanonical leading-zero masks).

The added autotest currently checks only the default `::/0` CLI output. It would also be valuable to configure a non-default source prefix and verify packets from both matching and non-matching source ranges.

